PR #85, opened 2015-01-26
[FEATURE] Implemented delivery status for products

christianbader:


Api/DeliveryStatus.php
```diff
@@ -0,0 +1,85 @@
...
+use JMS\Serializer\Annotation\VirtualProperty;
+use JMS\Serializer\Annotation\SerializedName;
+use JMS\Serializer\Annotation\ExclusionPolicy;
+use Sulu\Component\Rest\ApiWrapper;
+use Sulu\Bundle\ProductBundle\Entity\DeliveryStatusTranslation;
+
+/**
+ * The DeliveryStatus class which will be exported to the API
+ *
+ * @package Sulu\Bundle\ProductBundle\Api
```
dantleech: I think we want to exclude these annotations (`@pacakge`) as they are just repeating information (i.e. the Namespace). /cc @chirimoya @danrot

Api/DeliveryStatus.php
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public function __construct(Entity $type, $locale)
+    {
+        $this->entity = $type;
+        $this->locale = $locale;
+    }
+
+    /**
+     * The id of the type
+     * @return int The id of the type
```
dantleech:
````
/**
 * The id of the type
 *
 * @return int The ID of the field
 */
````

 blank line before return

File: Api/DeliveryStatus.php [hunk 3]
```diff
@@ -0,0 +1,85 @@
...
+     * @param string $name The name of the delivery status
+     */
+    public function setName($name)
+    {
+        $this->getTranslation()->setName($name);
+    }
+
+    /**
+     * Returns the translation for the current locale
+     *
```
dantleech: Missing return

Product/ProductManager.php
```diff
@@ -1126,6 +1141,22 @@ public function setStatusForProduct($product, $statusId)
...
     /**
+     * Sets the deliveryStatus for a given product
+     *
+     * @param Product $product
+     * @param int $statusId
+     * @throws Exception\ProductDependencyNotFoundException
+     */
+    public function setDeliveryStatusForProduct($product, $statusId)
+    {
+        $status = $this->deliveryStatusRepository->find($statusId);
```
notsure: could also be a try catch

PR #86, opened 2015-01-27
[FEATURE] Category ids are returned with each product as comma seperated sting

christianbader:


notsure:
ack

dantleech:
Why not return an array?

PR #87, opened 2015-01-28
[BUGFIX] fixed default delivery status and api entity

michaelzangerle:
__Tasks:__
- [x] done
- [ ] got feedback

__Informations:__

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tests pass?   | none
| Fixed tickets | none
| Doc           | none

### DIFF
--- a/Api/Product.php
+++ b/Api/Product.php
@@ -438,9 +438,12 @@ class Product extends ApiWrapper
      */
     public function getDeliveryStatus()
     {
-        if(!is_null($this->entity->getDeliveryStatus())){
-            return new DeliveryStatus($this->entity->getDeliveryStatus(), $this->locale);
+        $status = $this->entity->getDeliveryStatus();
+        if($status === NULL) {
+
+            return new DeliveryStatus($status, $this->locale);
         }
+
         return null;
     }
 
--- a/Api/Product.php
+++ b/Api/Product.php
@@ -439,7 +439,7 @@ class Product extends ApiWrapper
     public function getDeliveryStatus()
     {
         $status = $this->entity->getDeliveryStatus();
-        if($status === NULL) {
+        if($status === null) {
 
             return new DeliveryStatus($status, $this->locale);
         }
--- a/Api/Product.php
+++ b/Api/Product.php
@@ -438,7 +438,10 @@ class Product extends ApiWrapper
      */
     public function getDeliveryStatus()
     {
-        return new DeliveryStatus($this->entity->getDeliveryStatus(), $this->locale);
+        if(!is_null($this->entity->getDeliveryStatus())){
+            return new DeliveryStatus($this->entity->getDeliveryStatus(), $this->locale);
+        }
+        return null;
     }
 
     /**
--- a/Entity/DeliveryStatus.php
+++ b/Entity/DeliveryStatus.php
@@ -9,6 +9,10 @@ use Doctrine\ORM\Mapping as ORM;
  */
 class DeliveryStatus
 {
+
+    const AVAILABLE = 1;
+    const NOT_AVAILABLE = 2;
+
     /**
      * @var integer
      */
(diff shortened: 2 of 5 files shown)


Api/Product.php
```diff
@@ -438,7 +438,10 @@ public function setDeliveryStatus(DeliveryStatusEntity $deliveryStatus)
      */
     public function getDeliveryStatus()
     {
-        return new DeliveryStatus($this->entity->getDeliveryStatus(), $this->locale);
+        if(!is_null($this->entity->getDeliveryStatus())){
```
dantleech:
````
$deliveryStatus = $this->entitty->getDeliveryStatus();

if (null !== $deliveryStatus) {
    return new DeliveryStatus($deliveryStatus, $this->locale);
}
````

General standards tend towards using `null ===/!==` rather than `is_null` and you should assign method evaluations to variables rather than call them twice or more (it is less code).

File: Api/Product.php [hunk 2]
```diff
@@ -438,7 +438,10 @@ public function setDeliveryStatus(DeliveryStatusEntity $deliveryStatus)
...
+            return new DeliveryStatus($this->entity->getDeliveryStatus(), $this->locale);
+        }
+        return null;
```
dantleech: Always have an empty line before a `return`

Entity/DeliveryStatus.php
```diff
@@ -9,6 +9,10 @@
  */
 class DeliveryStatus
 {
+
```
dantleech: Remove extra line

Entity/DeliveryStatus.php
```diff
@@ -9,6 +9,10 @@
...
+    const AVAILABLE = 1;
+    const NOT_AVAILABLE = 2;
```
dantleech: hmm. why use integers? it makes debugging and inspecting the database a real pain. I know we do this everywhere else, but just saying. /cc @chirimoya @danrot @wachterjohannes

Product/ProductManager.php
```diff
@@ -926,6 +928,20 @@ public function save(
...
 
+        if (isset($data['deliveryStatus']) && isset($data['deliveryStatus']['id'])) {
+            $deliveryStatusId = $data['deliveryStatus']['id'];
+            /** @var DeliveryStatus $deliveryStatus */
+            $deliveryStatus = $this->deliveryStatusRepository->find($deliveryStatusId);
+            if (!$deliveryStatus) {
+                throw new ProductDependencyNotFoundException(self::$productDeliveryStatusClassEntityName, $deliveryStatusId);
+            }
+            $product->setTaxClass($deliveryStatus);
+        } elseif ($product->getDeliveryStatus() == null) {
```
dantleech: always compare null with `===`

Product/ProductManager.php
```diff
@@ -926,6 +928,20 @@ public function save(
             $product = $this->convertProduct($product, $publishedProduct);
         }
 
+        if (isset($data['deliveryStatus']) && isset($data['deliveryStatus']['id'])) {
```
dantleech: `isset($data['deliveryStatus]['id'])` is sufficient

````bash
$ echo "<?php var_dump(isset(\$foo['bar']['boo']));" | php
bool(false)
````

File: Product/ProductManager.php (hunk 5 above)
Comment (dantleech) [inline Product/ProductManager.php]: Also convention is `if (null === $var)`

````
$ symfony/ git grep "null ===" | wc -l
547
$ symfony git grep "=== null" | wc -l
11
````

File: Api/Product.php [hunk 7]
```diff
@@ -438,7 +438,13 @@ public function setDeliveryStatus(DeliveryStatusEntity $deliveryStatus)
      */
     public function getDeliveryStatus()
     {
-        return new DeliveryStatus($this->entity->getDeliveryStatus(), $this->locale);
+        $status = $this->entity->getDeliveryStatus();
+        if($status === NULL) {
```
dantleech: always lower case `null`

christianbader:
ack

PR #90, opened 2015-02-05
[BUGFIX] removed unused grunt clean task

notsure:


Gruntfile.js
```diff
@@ -21,22 +21,9 @@ module.exports = function (grunt) {
     grunt.initConfig({
         pkg: grunt.file.readJSON('package.json'),
         copy: {
```
danrot: Wouldn't you remove these empty objects also?

Gruntfile.js
```diff
@@ -21,22 +21,9 @@ module.exports = function (grunt) {
...
-            public: {
-                files: [
-                    {expand: true, cwd: 'Resources/public', src: ['**', '!**/scss/**'], dest: '../../../../../../web/bundles/suluproduct/'}
-                ]
-            }
         },
         clean: {
```
danrot: Same here, the options are not a separate task AFAIK

danrot:
The two issues are not that important, change it if you want. Otherwise ACK

PR #93, opened 2015-02-17
[FEATURE] Quickfix for product attributes

christianbader:


Product/ProductManager.php
```diff
@@ -769,6 +769,19 @@ public function save(
         }
 
         if (isset($data['attributes'])) {
+            // Delete attribures
```
turbo-ele: attribures?

Product/ProductManager.php
```diff
@@ -778,14 +791,24 @@ public function save(
...
-                    $productAttribute = new ProductAttribute();
-                    $attribute = $this->attributeRepository->find($attributeData['attributeId']);
-                    if (!$attribute) {
-                        throw new ProductDependencyNotFoundException(self::$attributeEntityName, $attributeId);
+                    $attributeIds[] = $attributeData['attributeId'];
+                    // Check if attribure exists for current product
+                    $productAttribute = $this->productAttributeRepository->findByAttributeIdAndProductId(
+                        $attributeData['attributeId'],
+                        $product->getId()
+                    );
```
turbo-ele: is that even possible after deleting all attributes of the product?

turbo-ele:
ack

PR #103, opened 2015-02-25
[FEATURE] added multiple delete of products

turbo-ele:


turbo-ele:
ack by @christianbader

PR #128, opened 2015-03-25
[FEATURE] Added function to product manager that returns all products with current...

christianbader:
... effective special prices assigned to it

### DIFF
--- a/Entity/SpecialPriceRepository.php
+++ b/Entity/SpecialPriceRepository.php
@@ -24,7 +24,7 @@ class SpecialPriceRepository extends EntityRepository
     {
         try {
             $qb = $this->createQueryBuilder('specialPrice')
-                ->add('where', ':now BETWEEN specialPrice.start AND specialPrice.end')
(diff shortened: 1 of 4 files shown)


Entity/SpecialPriceRepository.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Returns the current special prices
+     *
+     * @return SpecialPrice[]|null
+     */
+    public function findAllCurrent()
+    {
+        try {
+            $qb = $this->createQueryBuilder('specialPrice')
+                ->add('where', ':now BETWEEN specialPrice.start AND specialPrice.end')
```
turbo-ele: ->where(':now BETWEEN specialPrice.start AND specialPrice.end')

turbo-ele: instead of now and setting parameter you could use CURRENT_TIMESTAMP()

turbo-ele:
else ack

PR #137, opened 2015-04-20
[FEATURE] added scale price function, replaced input fields in PIM

notsure:


christianbader:
@turbo-ele fix tests!

christianbader:
Make sure that the tests are running without failure, then ack.

PR #169, opened 2015-06-03
[BUGFIX] is numeric check of id at product save

xandrej:
__Tasks:__
- [x] done
- [ ] todo
- [ ] got feedback

__Informations: product save bugfix

| Q            | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tests pass?  | yes
| Fixed tickets | https://github.com/massiveart/POOL-ALPIN/issues/1277
| Doc          | none

### DIFF
--- a/Resources/public/js/components/products/components/detail-form/main.js
+++ b/Resources/public/js/components/products/components/detail-form/main.js
@@ -97,7 +97,7 @@ define([
                 var data = this.sandbox.form.getData(formSelector),
                     supplierId;
 
-                if (!$.isNumeric(data.id)) {
+                if (!this.sandbox.dom.isAlphaNumeric(data.id)) {
                     delete data.id;
                 }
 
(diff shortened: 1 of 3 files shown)


michaelzangerle:
choose a meaningful PR title and use [FEATURE]/[BUGFIX]/...

Resources/public/js/components/products/components/detail-form/main.js
```diff
@@ -97,7 +97,7 @@ define([
                 var data = this.sandbox.form.getData(formSelector),
                     supplierId;
 
-                if (data.id === '') {
+                if (!$.isNumeric(data.id)) {
```
michaelzangerle: use the jquery abstraction

michaelzangerle: an id could also be a hash

turbo-ele:
unfortunately this does not solve the real issue and is just a quick fix, since the real problem is that the entity seems to return strange id's.

ACK anyway

PR #174, opened 2015-06-09
[BUGFIX] Fixed bug that causes an error when a new product without minimumOrderQuantity is added

christianbader:


turbo-ele:
ack

PR #178, opened 2015-06-11
[FEATURE] Added funtion to BaseProduct to check if it is a valid shop product

christianbader:


Entity/BaseProduct.php
```diff
@@ -1012,4 +1013,21 @@ public function getAreGrossPrices()
...
+    /**
+     * Helper method to check if the product is
+     * a valid shop product.
+     *
+     * @return bool
+     */
+    public function isValidShopProduct()
+    {
+        if ($this->getStatus()->getId() == Status::ACTIVE &&
+            $this->getPrices() && count($this->getPrices()) > 0
```
turbo-ele: check if getPrices function exists

christianbader: i will do it

PR #181, opened 2015-06-17
[FEATURE] Added index for internalItemNumber

turbo-ele:


turbo-ele:
ack by @christianbader

PR #189, opened 2015-07-10
[FEATURE] Added new units and mappings

christianbader:


DataFixtures/units.xml
```diff
@@ -307,4 +306,25 @@
...
+    </unit>
+    <unit>
+        <translations>
+            <de>Gebinde</de>
+            <en>Packages</en>
+        </translations>
+        <mappings>
+            <value>gebinde</value>
+            <value>geb</value>
+            <value>packages</value>
```
turbo-ele: pckgs?

turbo-ele:
ack

PR #196, opened 2015-07-20
[BUGFIX] removed default value for product cost

michaelzangerle:
__Tasks:__
- [x] done
- [ ] got feedback

__Informations:__

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tests pass?   | none
| Fixed tickets | https://github.com/massiveart/POOL-ALPIN/issues/1607
| Doc           | none

### DIFF
--- a/Entity/BaseProduct.php
+++ b/Entity/BaseProduct.php
@@ -51,7 +51,7 @@ abstract class BaseProduct implements ProductInterface
     /**
      * @var double
      */
-    protected $cost = 0;
+    protected $cost;
 
     /**
      * @var string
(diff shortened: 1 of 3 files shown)


Entity/BaseProduct.php
```diff
@@ -51,7 +51,7 @@
     /**
      * @var double
```
hacfi: maybe change this to `float` as well?

wachterjohannes: @michaelzangerle please change it here! we should stay with the phpdoc standard. also we should be avoid that this will not be regenerated sometime!

turbo-ele:
ack

PR #201, opened 2015-09-15
added pagerfanta to function findCurrentOfferedProducts

dmetzler1988:


Entity/SpecialPriceRepository.php
```diff
@@ -18,16 +20,24 @@ class SpecialPriceRepository extends EntityRepository
...
      * Returns the current special prices
      *
-     * @return SpecialPrice[]|null
+     * @param int $limit
+     * @param int $page
+     *
+     * @return null|Pagerfanta
      */
-    public function findAllCurrent()
+    public function findAllCurrent($limit = 1000, $page = 0)
```
turbo-ele: shouldn't `$page = 1` be the default?

turbo-ele:
ack

wachterjohannes:
@dmetzler1988 you have to be care of performance in case of a lot of products because pagerfanta uses the `Paginator` of doctrine and we had a lot of troubles with that ...

https://github.com/whiteoctober/Pagerfanta/blob/master/src/Pagerfanta/Adapter/DoctrineORMAdapter.php#L26

PR #206, opened 2015-10-28
[FEATURE] Added exception when product not active; cleanups

turbo-ele:
only return special prices of active products in ``findAllCurrent`` function

### DIFF
--- a/Product/Exception/ProductNotActiveException.php
+++ b/Product/Exception/ProductNotActiveException.php
@@ -13,15 +13,11 @@ namespace Sulu\Bundle\ProductBundle\Product\Exception;
 class ProductNotActiveException extends ProductException
 {
     /**
-     * The name of the entity.
-     *
      * @var string
      */
     private $entityName;
 
     /**
-     * The id of the object not found.
-     *
      * @var integer
      */
     private $id;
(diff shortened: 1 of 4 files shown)


Product/Exception/ProductNotActiveException.php
```diff
@@ -0,0 +1,59 @@
...
+{
+    /**
+     * The name of the entity.
+     *
+     * @var string
+     */
+    private $entityName;
+
+    /**
+     * The id of the object not found.
```
christianbader: ?

christianbader:
else ack

composer.json
```diff
@@ -31,7 +31,10 @@
         "symfony-cmf/testing": "~1.2",
         "sulu/document-manager": "dev-develop",
         "sulu/test-bundle": "dev-develop",
-        "massive/search-bundle": "0.7.*"
+        "massive/search-bundle": "dev-develop",
+        "pulse00/ffmpeg-bundle": "~0.5.2",
+        "php-ffmpeg/php-ffmpeg": "0.5.*",
```
hacfi: The two `ffmpeg` dependencies are not used at the moment.

danrot: @turbo-ele Why have you added them here?

turbo-ele: they 'have been" a dependency by sulu/sulu: "dev-develop"
"composer update" was not possible without that

danrot: Are you sure that's not only because of the search bundle?

turbo-ele: pretty sure, since composer told me that it cannot install since sulu/sulu needs ffmpeg to continue.

PR #207, opened 2015-10-30
[FEATURE] Added search terms to product

turbo-ele:


Product/ProductManager.php
```diff
@@ -1099,6 +1098,11 @@ public function save(
             $product->setCost(floatval($data['cost']));
         }
 
+        if (isset($data['searchTerms'])) {
+            $searchFields = $this->parseCommaSeparatedString($data['searchTerms']);
```
christianbader: searchTerms not searchFields

christianbader:
else ack

PR #217, opened 2015-11-20
added usage of price formatter

stemod:


turbo-ele:
ack

Api/Product.php
```diff
@@ -53,15 +54,26 @@ class Product extends ApiWrapper implements ApiProductInterface
     protected $accountManager;
 
     /**
+     * @var PriceFormatter
```
turbo-ele: no exclude needed here?

stemod: ExclusionPolicy is set to all here

PR #223, opened 2015-12-15
[ENHANCEMENT] Added category-translation to product-test-data

turbo-ele:


Tests/Resources/ProductTestData.php
```diff
@@ -84,7 +87,7 @@ class ProductTestData
     /**
      * @var Category
      */
-    private $category;
+    public $category;
```
stemod: do not use public property

stemod:
ack

PR #224, opened 2015-12-17
[FEATURE] Added pricing bundle support

turbo-ele:


Tests/Resources/ContactTestData.php
```diff
@@ -163,12 +177,9 @@ protected function createFixtures()
      *
      * @return AccountInterface
      */
-    public function createAccount($accountType = null)
+    public function createAccount($accountType = 0)
```
stemod: is there any constant for "0"?

stemod:
else ack

Tests/Resources/ContactTestData.php
turbo-ele: unfortunately, only in the bundle i don't want to reference

PR #225, opened 2015-12-17
[BUGFIX] bugfixed copying of search-terms on product copy

turbo-ele:


turbo-ele:
ack by @christianbader

PR #251, opened 2016-04-13
[BUGFIX] Removed default locale from function declaration

turbo-ele:
Implemented fallback for unit translation


Entity/Unit.php
```diff
@@ -50,128 +51,113 @@ public function setId($id)
     }
 
     /**
-     * Get id
-     *
-     * @return integer 
+     * @return integer
```
christianbader: int

christianbader:
ack

PR #254, opened 2016-04-21
[FEATURE] Created config for product locales

turbo-ele:


Product/ProductLocaleManager.php
```diff
@@ -0,0 +1,73 @@
...
+     * @param UserInterface $user
+     * @param null|string $requestLocale
+     *
+     * @return string
+     */
+    public function retrieveLocale(UserInterface $user, $requestLocale = null)
+    {
+        $languageMatch = null;
+        $userLanguage = explode('_', $user->getLocale());
+        $userLanguage = $userLanguage[0];
```
trickreich: in my opinion there's a better way doing this:
````
strstr($locale, '_', true)

File: Product/ProductLocaleManager.php [hunk 2]
```diff
@@ -0,0 +1,73 @@
...
+
+        // Use request locale if defined.
```
trickreich: move to first line of this method

Resources/public/js/components/products/main.js
```diff
@@ -66,6 +66,50 @@ define([
...
+     */
+    var retrieveDefaultLocale = function() {
+        var user = this.sandbox.sulu.user;
+        var productConfig = Config.get('sulu-product');
+        var defaultLocale = productConfig['fallback_locale'];
+        var locales = productConfig['locales'];
+        var languageMatch = null;
+
+        // Check if users locale contains localization.
+        var userLanguage = user.locale.split('_');
```
trickreich: maybe there's a better way to do this (string find with string replace)

Resources/public/js/components/products/main.js
```diff
@@ -407,18 +451,22 @@ define([
...
-         * Creates the view for the flat product list
+         * Creates the view for the flat product list.
          */
         renderList: function() {
             var $list = this.sandbox.dom.createElement('<div id="products-list-container"/>');
             this.html($list);
-            this.sandbox.start([
-                {name: 'products/components/list@suluproduct', options: {el: $list}}
-            ]);
+            this.sandbox.start([{
```
trickreich: maybe split array and object definition in separate lines

trickreich:
else ack

trickreich:
ack

PR #270, opened 2016-06-15
[BUGFIX] Small changes

trickreich:
Removed not used attribute types
Display fallback locale when adding a new attribute value to a product


turbo-ele:
Description could be a little bit better, like 'removed unused attribute fixtures'.

btw. it's never to late to start writing to CHANGELOG. i.m.o: now is the time ;)

else ack

CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+CHANGELOG for Sulu Product Bundle
+=================================
+
+* 0.10.3 (2016-06-15)
+    * ENHANCEMENT Removed not used attribute types
```
turbo-ele: .. from fixtures

Product/ProductManager.php
```diff
@@ -1892,15 +1892,20 @@ protected function processAttributes(array $data, ProductInterface $product, $lo
 
             // Add and change attributes.
             foreach ($data['attributes'] as $attributeData) {
-                if (!isset($attributeData['attributeId'])) {
+                if (!array_key_exists('attributeId', $attributeData) || !isset($attributeData['attributeId'])) {
```
turbo-ele: isset() returns false if !array_key_exists('attributeId', $attributeData)

turbo-ele:
else ack

PR #272, opened 2016-06-21
[ENHANCEMENT] Added method 'find' to StatusManager

trickreich:


Product/StatusManager.php
```diff
@@ -30,8 +30,8 @@ public function __construct(StatusRepository $statusRepository)
     }
 
     /**
-     * @param $locale
-     * @return null|Status[]
+     * @param string $locale
+     * @return null|self[]
```
turbo-ele: add blank line

turbo-ele: `@return null|[]`

Product/StatusManager.php
```diff
@@ -46,4 +46,14 @@ function (&$status) use ($locale) {
 
         return $statuses;
     }
+
+    /**
+     * @param integer $id
```
turbo-ele: int

Product/StatusManager.php
```diff
@@ -46,4 +46,14 @@ function (&$status) use ($locale) {
...
+     *
+     * @return null|self
```
turbo-ele: null|Status

Product/StatusManager.php
```diff
@@ -46,4 +46,14 @@ function (&$status) use ($locale) {
...
+     */
+    public function find($id)
```
turbo-ele: findById

turbo-ele:
else ack

PR #273, opened 2016-06-21
[ENHANCEMENT] Improved ContactTestData by adding use of ContactRepositories create …

turbo-ele:
…contact function


christianbader:
ack

PR #289, opened 2016-07-01
added delete addon by id action

stemod:


turbo-ele:
Add entry to changelog, else ack

PR #301, opened 2016-07-11
fixed parsing values

stemod:


turbo-ele:
are there any tests for fixtures import?
else ack

PR #302, opened 2016-07-11
added method to get products by category

stemod:


Product/ProductManager.php
```diff
@@ -881,6 +881,20 @@ public function findEntitiesByInternalItemNumber($internalItemNumber)
     }
 
     /**
+     * @param int $categoryId
+     *
```
turbo-ele: remove

turbo-ele:
else ack

PR #313, opened 2016-08-29
[ENHANCEMENT] Added config param 'display_recurring_prices'

trickreich:


Resources/views/Template/product.pricing.html.twig
```diff
@@ -60,6 +60,7 @@
                 </div>
             </div>
         </div>
+        {% if displayRecurringPrices == true %}
         <div class="grid-col-6">
```
sabinebaer: indent after if

Controller/TemplateController.php
```diff
@@ -135,13 +135,15 @@ public function productPricingAction(Request $request)
 
         $currencies = $this->getCurrencies($language);
         $defaultCurrency = $this->container->getParameter('sulu_product.default_currency');
+        $displayRecurringPrices = $this->container->getParameter('sulu_product.display_recurring_prices');
```
sabinebaer: "display" is a bit confusing here, as it does not only prevent displaying recurring prices somewhere in the frontend but disables the whole functionality.... So maybe enable_recurring_prices would be better?

sabinebaer:
else ACK

PR #317, opened 2016-09-02
[FEATURE] Moved price formatter form pricing to product bundle

turbo-ele:


Exception/PriceFormatterException.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+/*
```
trickreich: do we need this header now in all files of this bundle?

turbo-ele: yes, should be part of every file. i'm going to change that in every file that i touch.

Api/Product.php
```diff
@@ -33,14 +33,14 @@
...
 use Sulu\Bundle\ProductBundle\Entity\Type as TypeEntity;
 use Sulu\Bundle\ProductBundle\Entity\Unit as UnitEntity;
-use Sulu\Bundle\PricingBundle\Pricing\PriceFormatter;
+use Sulu\Bundle\ProductBundle\Util\PriceFormatter;
 use Sulu\Bundle\ProductBundle\Product\ProductLocaleManager;
 use Sulu\Component\Rest\ApiWrapper;
 use Sulu\Component\Security\Authentication\UserInterface;
 
 /**
  * The product class which will be exported to the API
```
trickreich: .

trickreich:
else ack

PR #320, opened 2016-09-05
[FEATURE] Product variant attributes

turbo-ele:
This PR implements a new relation between products and attributes: variantAttributes.

This relation is used for assigning attributes to a `Product with variants` that need to be defined by the variants.


trickreich:
ack

PR #322, opened 2016-09-06
[FEATURE] Added configuration for style-ci

turbo-ele:


Api/ApiProductInterface.php
```diff
@@ -10,7 +11,6 @@
 
 namespace Sulu\Bundle\ProductBundle\Api;
 
-
 interface ApiProductInterface
```
wachterjohannes: docs on each class/interface files

PR #323, opened 2016-09-06
[ENHANCEMENT] Renamed js variables

trickreich:


turbo-ele:
no changes on build?

also, do you want all current changes in develop in your next version? 

else ack

trickreich:
No changes in build files because I've just only renamed some variables.

Yes, finally we need all changes in the next version.

PR #331, opened 2016-10-18
[BUGFIX] Added own user settings key for product media list

trickreich:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Added own user settings key for product media list.


turbo-ele:
ack

PR #352, opened 2016-11-22
[BUGFIX] Fixed setting attributes of value 0

turbo-ele:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT

#### What's in this PR?

This PR fixes setting attributes of value '0'

#### Why?

Previously, when setting an attribute to a product with value '0', the relation was deleted. Now a value translation '0' is created.

### DIFF
--- a/Product/ProductManager.php
+++ b/Product/ProductManager.php
@@ -2009,7 +2009,7 @@ class ProductManager implements ProductManagerInterface
                 }
 
                 // If attribute value is empty do not add.
-                if (!$attributeDataValueName) {
+                if ('' === $attributeDataValueName || null === $attributeDataValueName) {
                     // If already set on product, remove attribute value translation.
                     if (array_key_exists($attributeId, $productAttributes)) {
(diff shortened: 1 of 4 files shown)


Product/ProductManager.php
```diff
@@ -2009,7 +2009,7 @@ protected function processAttributes(array $data, ProductInterface $product, $lo
                 }
 
                 // If attribute value is empty do not add.
-                if (!$attributeDataValueName) {
+                if ($attributeDataValueName === '') {
```
turbo-ele: the problem is that empty(0) or empty('0') will return true

wachterjohannes: shit but whats about `null`

turbo-ele: added null check

wachterjohannes:
ack

PR #357, opened 2017-01-19
[BUGFIX] Improved ProductLocaleManager to check if requested locale is available

trickreich:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no

#### What's in this PR?

Improved ProductLocaleManager to check if requested locale is available.


turbo-ele:
else ack

PR #368, opened 2017-10-13
fixed a bug in the addon overlay if price is overwritten with 0

mob8607:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Bugfix if a product has an addon and wants to overwrite the price with the value "0".

#### Why?

Currently this was stored, but not recognised correctly when opening the addon overlay again.
It didn't show that the price was overwritten at all.


sabinebaer:
ack

PR #377, opened 2018-05-07
Fixed deleting a product in the detail view

sabinebaer:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

The productId is passed correctly through all functions if a product is deleted from the product detail view.


Resources/public/js/components/products/util/header.js
```diff
@@ -85,11 +85,13 @@ define(['config'], function(Config) {
          *
          * @param {Object} sandbox
          * @param {Object} status
+         * @param {Integer} productId
```
trickreich: Integer isn't valid in javascript?

trickreich:
else ACK

PR #381, opened 2026-05-22
Add attribute admin

martinlagler:
This adds the Attribute entity with an own navigation point to Sulu.


config/doctrine/Product/Attribute.orm.xml
```diff
@@ -23,6 +23,9 @@
                      target-entity="Sulu\Product\Domain\Model\AttributeOptionInterface"
                      mapped-by="attribute">
             <cascade><cascade-persist/><cascade-remove/></cascade>
+            <order-by>
+                <order-by-field name="position" direction="ASC"/>
+            </order-by>
```
alexander-schranz: We should avoid this. It ends adding order by to every query selecting attributes. Which make it hard or sometimes impossible to optimize queries.

Orders should be done only by the queries we require them. Or I think in this case do it in the PHP getter method of getOptions method.

alexander-schranz (changes requested):
just stumbled over this in a quick look:

src/UserInterface/Controller/Admin/AttributeDetailsController.php
```diff
@@ -0,0 +1,206 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\HandleTrait;
+use Symfony\Component\Messenger\MessageBusInterface;
+
+/**
+ * @internal
+ */
+final class AttributeDetailsController implements SecuredControllerInterface
```
alexander-schranz: no need for `Details` postfix here

alexander-schranz (changes requested):
Can we cherry-pick the new CI and Tests unrelated to the Attributes in it's own PR to make this PR smaller and merging the tests and CI changes sooner.

src/Application/Message/CreateAttributeMessage.php
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Product\Application\Message;
+
+use Webmozart\Assert\Assert;
+
+class CreateAttributeMessage
```
alexander-schranz, line 31: this message should have `getName()`, `getDescription()`, ....

src/Application/Message/ModifyAttributeMessage.php
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Product\Application\Message;
+
+use Webmozart\Assert\Assert;
+
+class ModifyAttributeMessage
```
alexander-schranz, line 32: same as for create

src/Application/MessageHandler/CreateAttributeMessageHandler.php
```diff
@@ -0,0 +1,72 @@
...
+         *     name: string,
+         *     type: string,
+         *     description: string|null,
+         *     options: list<array{
+         *        type: string,
+         *        key: string,
+         *        name: string,
+         *     }>|null,
+         * } $data */
+        $data = $message->getData();
```
alexander-schranz: This service should never access `getData` the messages should have TypeSafe getter function for all core properties. We access.

`getData`should only be used to dispatch the `Event` class.

src/Application/MessageHandler/ModifyAttributeMessageHandler.php
```diff
@@ -0,0 +1,120 @@
...
+         *     type?: string,
+         *     name?: string,
+         *     description?: string|null,
+         *     options?: list<array{
+         *         type: string,
+         *         key: string,
+         *         name: string,
+         *     }>|null,
+         * } $data */
+        $data = $message->getData();
```
alexander-schranz: same the message is responsible to provide `getLocale()`, `getKey()`, `getOptions()` same for create

src/Application/MessageHandler/ModifyAttributeMessageHandler.php
```diff
@@ -0,0 +1,120 @@
...
+                $optionTranslation = $option->getTranslation($locale);
+                $optionName = $optionData['name'];
+
+                if (null !== $optionTranslation) {
+                    $optionTranslation->setName($optionName);
+                } else {
+                    $option->addTranslation(new AttributeOptionTranslation($option, $locale, $optionName));
+                }
+            }
+        }
```
alexander-schranz: Create a `AttributeMapper` service which you use in the Create and ModfiyHandler 

```php
    {
```

src/Application/MessageHandler/ModifyAttributeMessageHandler.php
```diff
@@ -0,0 +1,120 @@
...
+         * } $data */
+        $data = $message->getData();
+        $identifier = $message->getIdentifier();
+        $locale = $data['locale'];
+
+        $attribute = $this->attributeRepository->findOneBy(['uuid' => $identifier['uuid'] ?? null]);
+
+        if (null === $attribute) {
+            throw new AttributeNotFoundException(['uuid' => $identifier['uuid'] ?? null]);
+        }
```
alexander-schranz: 1. A modify message identifier is an array which can be anything which is unique filterable `['key' => ..]`, `['externalIdentifier' => ...]`, `['uuid' => ..]` or any combination which ends in uniquness.

```suggestion
        $attribute = $this->attributeRepository->getOneBy($message->getIdentifier());
```

src/Domain/Model/Attribute.php
```diff
@@ -84,6 +91,7 @@ public function setCurrentLocale(string $locale): self
     public function getTranslation(?string $locale = null): ?AttributeTranslationInterface
     {
         $locale ??= $this->currentLocale;
+
         foreach ($this->translations as $translation) {
```
alexander-schranz: we should use a criteria so only specific translation is loaded and not all:

```php
$criteria = Criteria::create()
    ->where(Criteria::expr()->eq('locale', $locale));

return $entity->getTranslations()
    ->matching($criteria)
    ->first() ?: null;
```

src/Domain/Repository/AttributeRepositoryInterface.php
```diff
@@ -0,0 +1,28 @@
...
+namespace Sulu\Product\Domain\Repository;
+
+use Sulu\Product\Domain\Model\AttributeInterface;
+
+interface AttributeRepositoryInterface
+{
+    public function create(): AttributeInterface;
+
+    /** @param array<string, mixed> $criteria */
+    public function findOneBy(array $criteria): ?AttributeInterface;
```
alexander-schranz: add also `getOneBy` which throws a AttributeNotFoundException if null e.g.:

```suggestion
    /** @param array<string, mixed> $criteria */
    public function findOneBy(array $criteria): ?AttributeInterface;

    /**
     * @throws AttributeNotFoundExcpetion
     */
    public function getOneBy(array $criteria): AttributeInterface;
```

src/Domain/Repository/AttributeRepositoryInterface.php
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Sulu\Product\Domain\Repository;
+
+use Sulu\Product\Domain\Model\AttributeInterface;
+
+interface AttributeRepositoryInterface
+{
+    public function create(): AttributeInterface;
+
+    /** @param array<string, mixed> $criteria */
```
alexander-schranz: we should specificy `criteria` also this should be named `filters`.

```php
@param array{
   uuid?: string,
   externalIdentifier?: string,
   key?: string,
   // whatever filters we have in core, ...
} $filters
```

alexander-schranz (changes requested):

src/Application/MessageHandler/CreateAttributeMessageHandler.php
```diff
@@ -13,57 +13,23 @@
...
-use Sulu\Product\Domain\Model\AttributeTranslation;
 use Sulu\Product\Domain\Repository\AttributeRepositoryInterface;
 
 final class CreateAttributeMessageHandler
 {
-    public function __construct(private AttributeRepositoryInterface $attributeRepository)
-    {
+    public function __construct(
+        private AttributeRepositoryInterface $attributeRepository,
+        private AttributeMapper $attributeMapper,
```
alexander-schranz, line 25: we should create a `AttributeMapperInterface` which is later the extension point for all who extends the attribute with new logic and so we inject here `iterable<AttributeMapperInterface>` but currently fine to merge this to move forward.

alexander-schranz (approved):

PR #386, opened 2026-06-08
Add attribute group

martinlagler:


config/doctrine/Product/Attribute.orm.xml
```diff
@@ -25,5 +28,11 @@
                      mapped-by="attribute">
             <cascade><cascade-persist/><cascade-remove/></cascade>
         </one-to-many>
+
+        <many-to-one field="attributeGroup"
+                     target-entity="Sulu\Product\Domain\Model\AttributeGroupInterface"
+                     fetch="EAGER">
```
alexander-schranz: We should not eager load this relation. We only require the id not the whole entity so the Entity is not needed to be loaded.

config/doctrine/Product/Attribute.orm.xml
```diff
@@ -25,5 +28,11 @@
                      mapped-by="attribute">
             <cascade><cascade-persist/><cascade-remove/></cascade>
         </one-to-many>
+
+        <many-to-one field="attributeGroup"
```
alexander-schranz: I would not prefix this with `attribute` as like I do have `$attriubte->getTranslation()` I would go with `$attribute->getGroup()`.

config/doctrine/Product/AttributeGroup.orm.xml
```diff
@@ -18,7 +18,8 @@
 
         <one-to-many field="groupAttributes"
                      target-entity="Sulu\Product\Domain\Model\AttributeGroupAttributeInterface"
-                     mapped-by="attributeGroup">
+                     mapped-by="attributeGroup"
+                     orphan-removal="true">
             <cascade><cascade-persist/><cascade-remove/></cascade>
```
alexander-schranz: never use `cascade-remove` also always multiline this:

```
            <cascade>
                <cascade-persist/><cascade-remove/>
            </cascade>
```

config/forms/Attribute/attribute_details.xml
```diff
@@ -40,6 +40,18 @@
             </params>
         </property>
 
+        <property name="attributeGroup" type="single_attribute_group_selection" mandatory="true" colspan="9">
```
alexander-schranz: would just call it `group`

src/Domain/Model/AttributeGroup.php
```diff
@@ -106,7 +106,7 @@ public function removeTranslation(AttributeGroupTranslationInterface $translatio
     public function getGroupAttributes(): array
     {
         $groupAttributes = $this->groupAttributes->toArray();
-        \usort($groupAttributes, static fn (AttributeGroupAttributeInterface $a, AttributeGroupAttributeInterface $b) => $a->getPosition() <=> $b->getPosition());
+        \usort($groupAttributes, static fn (AttributeGroupAttributeInterface $a, AttributeGroupAttributeInterface $b) => $a->getAttribute()->getPosition() <=> $b->getAttribute()->getPosition());
```
alexander-schranz: make this readable put it into multiple lines:

```php
        \usort(
            $groupAttributes,
            static fn (AttributeGroupAttributeInterface $a, AttributeGroupAttributeInterface $b)
                => $a->getAttribute()->getPosition() <=> $b->getAttribute()->getPosition(),
        );
```

src/Domain/Model/AttributeInterface.php
```diff
@@ -58,4 +58,12 @@ public function getOption(string $key): ?AttributeOptionInterface;
...
 
     public function removeOption(AttributeOptionInterface $option): self;
+
+    public function getPosition(): int;
+
+    public function setPosition(int $position): self;
+
+    public function getAttributeGroup(): AttributeGroupInterface;
+
+    public function setAttributeGroup(AttributeGroupInterface $attributeGroup): self;
```
alexander-schranz: getGroup and setGroup

src/Domain/Repository/AttributeGroupRepositoryInterface.php
```diff
@@ -22,6 +22,9 @@ public function create(): AttributeGroupInterface;
     /** @param array<string, mixed> $criteria */
     public function findOneBy(array $criteria): ?AttributeGroupInterface;
 
+    /** @param array<string, mixed> $criteria */
+    public function countGroupAttributes(array $criteria): int;
```
alexander-schranz: add `countBy(array $filters)`

@Prokyonn we did name it Filters or in the new repos not criteria?

src/Domain/Repository/AttributeRepositoryInterface.php
```diff
@@ -13,15 +13,28 @@
...
+
+    /**
+     * @return AttributeInterface[]
+     */
+    public function findByGroupWithPositionAtLeast(AttributeGroupInterface $group, int $position, ?AttributeInterface $exclude = null): array;
+
+    /**
+     * @return AttributeInterface[]
+     */
+    public function findByGroupWithPositionBetween(AttributeGroupInterface $group, int $min, int $max, ?AttributeInterface $exclude = null): array;
```
alexander-schranz, line 60: We should move this into an own Doctrine Listener and not do it inside the repo.

To move forward add a `@internal` and a `TODO`.

src/Infrastructure/Doctrine/Repository/AttributeGroupRepository.php
```diff
@@ -49,6 +50,19 @@ public function findOneBy(array $criteria): ?AttributeGroupInterface
         return $attributeGroup;
     }
 
+    public function countGroupAttributes(array $criteria): int
```
alexander-schranz: This should no be on this repository instead use in the `AttributeRepository::countBy(['group' => $groupUuid])` inside the service

src/Domain/Model/Attribute.php
```diff
@@ -36,10 +36,15 @@ class Attribute implements AttributeInterface
     /** @var Collection<int, AttributeOptionInterface> */
     protected Collection $options;
 
-    public function __construct()
+    protected int $position = 0;
+
+    protected AttributeGroupInterface $attributeGroup;
```
alexander-schranz:
```
 protected AttributeGroupInterface $group;
```

src/Domain/Model/Attribute.php
```diff
@@ -165,4 +170,28 @@ public function removeOption(AttributeOptionInterface $option): self
...
+    {
+        return $this->attributeGroup;
+    }
+
+    public function setAttributeGroup(AttributeGroupInterface $attributeGroup): self
+    {
+        $this->attributeGroup = $attributeGroup;
+
+        return $this;
+    }
```
alexander-schranz, line 186: getGroup and setGroup

alexander-schranz (changes requested):

src/Application/Mapper/AttributeMapper.php
```diff
@@ -15,20 +15,76 @@
...
+
     public function mapAttributeData(AttributeInterface $attribute, CreateAttributeMessage|ModifyAttributeMessage $message): void
     {
         $attribute->setKey($message->getKey());
-        $attribute->setType($message->getType());
-
         $this->mapTranslation($attribute, $message);
         $this->mapOptions($attribute, $message);
+
+        if ($message instanceof CreateAttributeMessage) {
```
alexander-schranz, line 37: as discussed we may will move this into a doctrine listener. @martinlagler add a todo for it.

src/Domain/Model/AttributeGroup.php
```diff
@@ -0,0 +1,132 @@
...
+namespace Sulu\Product\Domain\Model;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+
+class AttributeGroup implements AttributeGroupInterface
+{
+    protected int $id;
+
+    protected ?string $uuid = null;
```
alexander-schranz, line 23: we should discuss if there is really an advantage to keep here both (id, uuid)? Think for article, pages and snippets we only used `uuid` and used that as `id`. I'm unsure if its worth the confusen and go fully with `uuid` only. Also easier if we ever need to sync between 2 systems.

@Prokyonn what do you think? /cc @chirimoya

alexander-schranz (commented):
We should think of splitting the `Message` and `MessageHandler` namespaces into `Attribute`, `AttributeGroup`, `Product`, ... but can be done later.

alexander-schranz:
Merging this as it is to move forward :)

src/Domain/Model/AttributeGroup.php
Prokyonn: If we use uuid v7 I think it is fine to stick with only uuid. The insertion performance should be similar between simple id and uuid v7.

PR #387, opened 2026-06-10
Add measurement + min/max value to attributes

martinlagler:


config/doctrine/Product/Attribute.orm.xml
```diff
@@ -8,10 +8,14 @@
         </id>
 
         <field name="uuid" type="guid" nullable="true" unique="true"/>
+        <field name="externalIdentifier" type="string" length="255" nullable="true" column="external_identifier"/>
         <field name="key" type="string" length="255" nullable="false" column="attribute_key" unique="true"/>
         <field name="type" type="string" length="32" nullable="false">
             <options><option name="default">number</option></options>
         </field>
+        <field name="measurementFamily" type="string" length="64" nullable="true" column="measurement_family"/>
+        <field name="unit" type="string" length="64" nullable="true"/>
```
alexander-schranz: For me `measurementFamily` and `unit` are type specific (number) configs and so I would store them in the `config` value.

@Prokyonn what do you think?

Prokyonn: As long as we do not want to sort/filter with these values in the sulu list, then I would also move them to the config 👍

src/Domain/Model/Attribute.php
```diff
@@ -104,24 +100,36 @@ public function setType(string $type): self
 
     public function getMeasurementFamily(): ?string
```
alexander-schranz: We should only have a `setConfig` and `getConfig` and use inside the form xml `config/measurementFamily`, `config/unit`, so you not need care about this.

config/forms/Attribute/attribute_details.xml
```diff
@@ -0,0 +1,108 @@
...
+                <title>sulu_product.min</title>
+            </meta>
+        </property>
+
+        <property name="max" type="number" colspan="6"
+                  visibleCondition='type == "number"'>
+            <meta>
+                <title>sulu_product.max</title>
+            </meta>
+        </property>
```
alexander-schranz: Lets make a concept for making this form extendable per Type. 

So the `NumberAttributeType` provides the form data for this. But to move forward here currently hardcode it with `config/..` inside this XML

alexander-schranz (changes requested):

alexander-schranz (approved):

alexander-schranz:
We sure need to discuss how we want to configure the measurement families and where to put such Service, I already had some discussion with @Prokyonn about it. But lets move forward here.

PR #388, opened 2026-06-17
Add product family base

martinlagler:


config/doctrine/Product/Product.orm.xml
```diff
@@ -13,6 +13,12 @@
         <field name="code" type="string" nullable="true" unique="true" length="255"/>
         <field name="externalIdentifier" type="string" length="255" nullable="true" column="external_identifier"/>
 
+        <many-to-one field="productFamily"
+                     target-entity="Sulu\Product\Domain\Model\ProductFamilyInterface">
+            <join-column name="idProductFamilies" referenced-column-name="id"
+                         nullable="true" on-delete="SET NULL"/>
+        </many-to-one>
```
alexander-schranz, line 20: I think the productFamily can not be `null`

```suggestion
        <many-to-one field="productFamily"
                     target-entity="Sulu\Product\Domain\Model\ProductFamilyInterface">
            <join-column name="idProductFamilies" referenced-column-name="id"/>
        </many-to-one>
```

We should throw an exception when a family is tried to be removed which still have products.

src/Domain/Model/Product.php
```diff
@@ -83,6 +85,18 @@ public function setExternalIdentifier(?string $externalIdentifier): self
         return $this;
     }
 
+    public function getProductFamily(): ?ProductFamilyInterface
```
alexander-schranz: think the family is a requried constructor parameter not being nullable in our case.

src/UserInterface/Controller/Admin/ProductFamilyController.php
```diff
@@ -0,0 +1,212 @@
...
+
+            $attributeId = (int) $matches[1];
+            $familyAttributes[] = [
+                'attribute' => $attributeId,
+                'required' => (bool) $request->request->get('attribute_' . $attributeId . '_required', false),
+            ];
+        }
+
+        return $familyAttributes;
+    }
```
alexander-schranz, line 192: I think we can use `attributes/<id>/enabled` and `attributes/<id>/required` in the form so we not need this methods to convert things in the controller.

src/UserInterface/Controller/Admin/ProductFamilyController.php
```diff
@@ -0,0 +1,212 @@
...
+        ];
+
+        foreach ($family->getFamilyAttributes() as $familyAttribute) {
+            $attributeId = $familyAttribute->getAttribute()->getId();
+            $data['attribute_' . $attributeId . '_enabled'] = true;
+            $data['attribute_' . $attributeId . '_required'] = $familyAttribute->isRequired();
+        }
+
+        return $data;
+    }
```
alexander-schranz, line 223: We should handle this over the `xml` that you already put this into a format like `attributes/<attribute_uuid>/enabled` and `attributes/<attribute_uuid>/required`.

src/Application/MessageHandler/ModifyProductFamilyMessageHandler.php
```diff
@@ -0,0 +1,83 @@
...
+
+            $attribute = $this->attributeRepository->findOneBy(['id' => $attributeId]);
+            if (null === $attribute) {
+                continue;
+            }
+
+            $familyAttribute = new ProductFamilyAttribute($family, $attribute);
+            $familyAttribute->setRequired($entry['required']);
+            $family->addFamilyAttribute($familyAttribute);
+        }
```
alexander-schranz: we should put this also in a ProductFamilyMapper which we use in create and modify message

alexander-schranz (changes requested):

src/Application/MessageHandler/ModifyProductFamilyMessageHandler.php
```diff
@@ -0,0 +1,39 @@
...
+use Sulu\Product\Application\Mapper\ProductFamilyMapperInterface;
+use Sulu\Product\Application\Message\ModifyProductFamilyMessage;
+use Sulu\Product\Domain\Model\ProductFamilyInterface;
+use Sulu\Product\Domain\Repository\ProductFamilyRepositoryInterface;
+
+final class ModifyProductFamilyMessageHandler
+{
+    public function __construct(
+        private ProductFamilyRepositoryInterface $productFamilyRepository,
+        private ProductFamilyMapperInterface $productFamilyMapper,
```
alexander-schranz, line 25: Mappers are extension point via tagged services.

```suggestion
        /** @var iterable<ProductFamilyMapperInterface> */
        private iterable $productFamilyMapper,
```

src/Domain/Model/ProductInterface.php
```diff
@@ -49,11 +49,15 @@ public function addTranslation(ProductTranslationInterface $translation): self;
     public function removeTranslation(ProductTranslationInterface $translation): self;
 
     /**
-     * @return Collection<int, ProductAttributeInterface>
+     * @return Collection<int, ProductAttributeValueInterface>
      */
     public function getAttributes(): Collection;
```
alexander-schranz, line 54: @Prokyonn what do you think, should we keep here Attribute or go with `getAttributeValues()` as we decided to go with `ProductAttributeValue`.

src/Domain/Repository/AttributeRepositoryInterface.php
```diff
@@ -21,6 +21,8 @@
  * @phpstan-type AttributeRepositoryFilters array{
  *     uuid?: string,
  *     key?: string,
+ *     id?: int,
```
alexander-schranz, line 24: we need discuss if we keep only uuid and throw id away or not 🤔

src/UserInterface/Controller/Admin/ProductFamilyController.php
```diff
@@ -0,0 +1,224 @@
...
+                'enabled' => (bool) ($entry['enabled'] ?? false),
+                'required' => (bool) ($entry['required'] ?? false),
+            ];
+        }
+
+        return $attributes;
+    }
+
+    /** @return array<string, mixed> */
+    private function serializeProductFamily(ProductFamilyInterface $family, string $locale): array
```
alexander-schranz, line 195: don't need be tackled now but for extendability of the entities we need probably use a serializer.

alexander-schranz:
@martinlagler keep the comments in mind for upcoming PR.

src/Domain/Model/ProductInterface.php
Prokyonn: I would go with `getAttributeValues`or even `getProductAttributeValues`because then it makes more sense to get the attribute via `getProductAttributeValue()[0]->getAttribute()` with the current naming it would bet `getAttribute()[0]->getAttribute()`

PR #389, opened 2026-06-23
Add product family relation

martinlagler:


config/forms/Attribute/attribute_details.xml
```diff
@@ -86,7 +86,7 @@
             </meta>
         </property>
 
-        <property name="description" type="text_editor">
+        <property name="description" type="text_area">
```
alexander-schranz: Why text area here? Think we already use a text editor for Sulu Core Category Description then I would use it also here?

For usage in Sulu XML I would call `strip_tags` currently in the Form Metadata Visitor :)

martinlagler: Because of the xml description, okay i can use strip_tags

martinlagler:
I’ll merge this PR, as the following PR will do some major refactoring.

PR #390, opened 2026-06-24
Feature/auditable

martinlagler:


src/UserInterface/Controller/Admin/ProductDetailsController.php
```diff
@@ -164,6 +164,33 @@ public function getLocale(Request $request): string
         return $request->query->getString('locale', $request->getLocale());
     }
 
+    /**
+     * @return CreateProductMessageData
```
Prokyonn: docblock should be removed

src/UserInterface/Controller/Admin/AttributeController.php
```diff
@@ -168,6 +168,33 @@ private function getData(Request $request): array
     }
 
     /** @return array<string, mixed> */
+    /**
```
Prokyonn: here is also something weird happening with the docblock :D

src/UserInterface/Controller/Admin/AttributeController.php
```diff
@@ -168,6 +168,33 @@ private function getData(Request $request): array
...
+     * @param array<mixed> $rows
+     *
+     * @return array<int, array<string, mixed>>
+     */
+    private static function normalizeDateTimes(array $rows): array
```
Prokyonn: This method is copied into 4 controllers, I think we should do the serialization the same way is in the other sulu 3.0 controllers like ArticleController:

```
 return new JsonResponse($this->normalizer->normalize(
     $listRepresentation->toArray(),
     'json',
 ));
```

Prokyonn (changes requested):

Prokyonn (approved):

PR #391, opened 2026-06-26
Add Versioning to product and add localized attribute, Refactoring product 

martinlagler:


Prokyonn (commented):
Automated review pass (correctness + cleanup). Most of the refactor verified clean — the trash store/restore round-trip, the `existBy` relocation into `ProductDetailsDataMapper`, and the getter renames all check out. Inline comments cover the specific findings.

Three more findings touch code this PR does not modify, so they cannot be anchored inline:

- **`ModifyAttributeGroupMessageHandler`** — `defaultLocale` is only set in `CreateAttributeGroupMessageHandler`, never backfilled on modify (unlike `AttributeMapper`/`ProductFamilyMapper`, which set it in the shared create+modify path). A group with `defaultLocale = null` never gets it set on edit, so the new default-locale translation fallback never applies and its name stays blank in other locales.
- **`src/Application/Mapper/ProductMapperInterface.php`** — now dead code: every implementor (`ProductContentMapper`, `ProductDetailsMapper`) was deleted in this PR and there are no remaining references. Safe to delete the file.
- **`ProductController::getCreateData()` / `getModifyData()`** — byte-identical, and the `template` default block is duplicated across `CreateProductMessageHandler` and `ModifyProductMessageHandler`. Worth collapsing each into one shared helper.

src/Infrastructure/Symfony/HttpKernel/SuluProductBundle.php
```diff
@@ -922,7 +919,7 @@ public function prependExtension(ContainerConfigurator $container, ContainerBuil
                         ],
                         ProductDimensionContentInterface::RESOURCE_KEY => [
                             'routes' => [
-                                'detail' => 'sulu_product.get_product_content',
+                                'detail' => 'sulu_product.get_product',
```
Prokyonn, line 936: This correctly repoints the `RESOURCE_KEY` detail route to `sulu_product.get_product`, but the same fix is missing just below: the `products_versions` resource (line 928) still declares `'detail' => 'sulu_product.get_product_content'`, and that route was **removed in this PR** — it no longer exists anywhere in the codebase. Resolving the versions detail link will throw `RouteNotFoundException` (500). It should be `sulu_product.get_product` here too.

src/UserInterface/Controller/Admin/AttributeController.php
```diff
@@ -182,6 +182,10 @@ private function serializeAttribute(?AttributeInterface $attribute, string $loca
         }
 
         $translation = $attribute->getTranslation($locale);
+        if (null === $translation) {
+            $defaultLocale = $attribute->getDefaultLocale();
+            $translation = null !== $defaultLocale ? $attribute->getTranslation($defaultLocale) : null;
```
Prokyonn, line 187: `serializeAttribute()`'s return array (just below) never emits `localized`, but the form binds a `localized` checkbox and `AttributeMapper::mapAttributeData()` applies it on **every** modify (not just create, unlike `type`). So editing an existing `localized=true` attribute loads the box unchecked and saving flips it to `false`, which then misroutes its values to the unlocalized dimension content. Please add `'localized' => $attribute->isLocalized()` to the returned array.

src/Infrastructure/Doctrine/Repository/ProductRepository.php
```diff
@@ -128,17 +127,30 @@ public function existBy(array $filters): bool
 
         $code = $filters['code'] ?? null;
         if (null !== $code) {
-            $queryBuilder->andWhere('product.code = :code')
+            $queryBuilder
+                ->innerJoin('product.dimensionContents', 'pdcCode')
+                ->andWhere('pdcCode.code = :code')
+                ->andWhere('pdcCode.locale IS NULL')
```
Prokyonn, line 133: This uniqueness join scopes only to `locale IS NULL`, missing the `stage = 'draft' AND version = 0` scoping that `config/lists/products.xml` uses. Since `code` now lives on per-stage/version dimension-content rows, a stale `code` on a live/historical row makes `existBy` report a code as taken after it has been changed+published — legitimate reuse then fails with `ProductCodeNotUniqueException`. The `productFamilyUuid` join just below has the same gap. Both need the draft/current-version filter.

src/Infrastructure/Doctrine/Repository/ProductFamilyRepository.php
```diff
@@ -101,8 +101,15 @@ public function createQueryBuilder(array $filters): QueryBuilder
...
-            $queryBuilder->innerJoin(Product::class, 'product', Join::WITH, 'product.productFamily = productFamily')
-                ->andWhere('product.uuid = :productUuid')
+            $queryBuilder
+                ->innerJoin(
+                    ProductDimensionContentInterface::class,
+                    'pdc',
+                    Join::WITH,
+                    'pdc.productFamily = productFamily',
+                )
+                ->andWhere('pdc.locale IS NULL')
```
Prokyonn: Same missing stage/version scope as `existBy`, and `findOneBy()`/`getOneBy()` only catch `NoResultException`. When a product's unlocalized dimension contents reference more than one family across a draft/live/version boundary, `getSingleResult()` throws an uncaught `NonUniqueResultException` and 500s the attribute form (`ProductAttributeFormMetadataVisitor::visitFormMetadata`). Scoping this join to the current draft (`stage = 'draft' AND version = 0`) makes only one row match.

src/Infrastructure/Symfony/Serializer/Normalizer/ProductNormalizer.php
```diff
@@ -35,46 +29,8 @@ public function __construct(
...
-            if (!$this->attributeTypeRegistry->has($attribute->getType())) {
-                continue;
-            }
-
-            $type = $this->attributeTypeRegistry->get($attribute->getType());
-            $attributes[$attribute->getId()] = $type->readValue($value);
-        }
-
         return [
             'id' => $data->getUuid(),
```
Prokyonn, line 33: `normalize()` now returns only `id`, so the `POST /products` 201 response drops title/code/family/attributes while `PUT`/`GET` return the full merged content — any consumer reading the create response gets just the id. If this is intentional (the entity no longer exposes those getters), a short note would help; otherwise consider mirroring `putAction` and returning the resolved content.

src/Infrastructure/Sulu/Content/DataMapper/ProductDetailsDataMapper.php
```diff
@@ -0,0 +1,73 @@
...
+            $unlocalizedDimensionContent->setCode($code);
+        }
+
+        if (\array_key_exists('externalIdentifier', $data)) {
+            /** @var string|null $externalIdentifier */
+            $externalIdentifier = $data['externalIdentifier'];
+            $unlocalizedDimensionContent->setExternalIdentifier($externalIdentifier);
+        }
+
+        if (\array_key_exists('productFamily', $data) && null !== $data['productFamily']) {
```
Prokyonn, line 66: `productFamily` is applied only when the key exists **and** is non-null, and `ProductDimensionContent::setProductFamily()` is non-nullable despite the column/getter being nullable — so `productFamily: null` (a clear request) is silently ignored. Not reachable via the UI (field is `disabledCondition="!!id"`), but a silent no-op on the API. Consider making the setter nullable and applying `null` explicitly.

src/Infrastructure/Sulu/Content/DataMapper/ProductDetailsDataMapper.php
```diff
@@ -0,0 +1,73 @@
...
+            && \array_key_exists('title', $data)
+        ) {
+            /** @var string|null $title */
+            $title = $data['title'];
+            $localizedDimensionContent->setTitle($title);
+        }
+
+        if (\array_key_exists('code', $data)) {
+            /** @var string|null $code */
+            $code = \is_string($data['code']) ? $data['code'] : null;
```
Prokyonn, line 50: `\is_string(...) ? ... : null` silently coerces a non-string `code` (e.g. a numeric JSON value) to `null`, wiping the stored code with no error. Consider rejecting non-string input rather than clearing the value.

src/Application/Mapper/AttributeMapper.php
```diff
@@ -32,6 +32,16 @@ public function mapAttributeData(AttributeInterface $attribute, CreateAttributeM
     {
         $attribute->setKey($message->getKey());
         $attribute->setConfig($message->getConfig());
+
+        $data = $message->getData();
+        if (\array_key_exists('localized', $data)) {
+            $attribute->setLocalized((bool) $data['localized']);
```
Prokyonn, line 38: `(bool) $data['localized']` treats the string `"false"` as `true` (any non-empty string is truthy), so a client serializing booleans as strings would silently mark the attribute localized. `filter_var($data['localized'], FILTER_VALIDATE_BOOL)` is safer.

src/Infrastructure/Symfony/HttpKernel/SuluProductBundle.php
```diff
@@ -323,38 +319,51 @@ public function loadExtension(array $config, ContainerConfigurator $container, C
...
 
-        $services->set('sulu_product.product_attribute_value_mapper')
-            ->class(ProductAttributeValueMapper::class)
+        $services->set('sulu_product.product_attributes_data_mapper')
+            ->class(ProductAttributesDataMapper::class)
             ->args([
                 new Reference('sulu_product.attribute_type_registry'),
             ])
-            ->tag('sulu_product.product_mapper');
+            ->tag('sulu_content.data_mapper');
```
Prokyonn: `ProductAttributesDataMapper` early-returns when `getProductFamily()` is null, so it depends on `product_details_data_mapper` running first — but both are tagged `sulu_content.data_mapper` with no `priority` (the merger above sets one explicitly). It works only because of registration order; a reorder or a third-party priority would silently drop all submitted attribute values on create. Please set explicit priorities so details runs before attributes.

config/doctrine/Product/ProductDimensionContent.orm.xml
```diff
@@ -29,5 +29,26 @@
                 <cascade-persist/>
             </cascade>
         </one-to-many>
+
+        <field name="code" type="string" length="255" nullable="true"/>
```
Prokyonn, line 37: Moving `code` here dropped the old `unique="true"` from `Product.code`, so uniqueness is now enforced solely by the app-level `existBy` check (which also has the scoping bug noted in `ProductRepository`). Worth confirming the intended enforcement — as-is there is no DB-level guard and the app-level one is race-prone.

config/doctrine/Product/Attribute.orm.xml
```diff
@@ -21,6 +21,12 @@
 
         <field name="config" type="json" nullable="false"/>
 
+        <field name="defaultLocale" type="string" length="5" nullable="true" column="default_locale"/>
+
+        <field name="localized" type="boolean" nullable="false">
+            <options><option name="default">false</option></options>
```
Prokyonn: Boolean default uses `false` where the rest of the codebase uses the numeric form (see `ProductFamilyAttribute.orm.xml`'s `required` -> `0`). `false` risks cross-platform DDL differences / perpetual schema-diff noise — suggest `<option name="default">0</option>`.

src/Infrastructure/Sulu/Admin/ProductAttributeFormMetadataVisitor.php
```diff
@@ -79,7 +88,8 @@ public function visitFormMetadata(FormMetadata $formMetadata, string $locale, ar
                 continue;
             }
 
-            $translation = $attribute->getTranslation($locale);
+            $translation = $attribute->getTranslation($locale)
```
Prokyonn, line 91: This `getTranslation($locale) ?? getTranslation($defaultLocale)` fallback is repeated ~7x (here, `ProductFamilyFormMetadataVisitor`, `ProductFamilyNormalizer`, `AttributeController` x2, `AttributeGroupController`) and the `defaultLocale` bootstrap 3x. A shared `getTranslationOrDefault()` on the model/trait would remove the duplication and keep the fallback behaviour consistent.

config/doctrine/Product/ProductDimensionContent.orm.xml
Prokyonn: @martinlagler maybe it makes sense to create a unique index for code/locale/stage/version, then we would ensure this again on database level

src/Application/MessageHandler/CreateProductMessageHandler.php
```diff
@@ -61,15 +60,18 @@ public function __invoke(CreateProductMessage $message): ProductInterface
...
 
-        $productFamily = $this->productFamilyRepository->getOneBy(['uuid' => $message->getProductFamily()]);
+        $product = $this->productRepository->createNew($message->getUuid());
 
-        $product = $this->productRepository->createNew($productFamily, $message->getUuid());
+        $this->productRepository->add($product);
 
-        foreach ($this->productMappers as $productMapper) {
-            $productMapper->mapProductData($product, $data);
+        if (!\array_key_exists('template', $data)) {
```
Prokyonn, line 66: How can this be empty?

src/Application/MessageHandler/ModifyProductMessageHandler.php
```diff
@@ -41,32 +37,25 @@ public function __invoke(ModifyProductMessage $message): ProductInterface
...
+                    'locale' => $locale,
+                    'stage' => [DimensionContentInterface::STAGE_DRAFT, DimensionContentInterface::STAGE_LIVE],
                 ],
-            ]
+            ]],
         );
 
-        if (null !== $code && $code !== $product->getCode() && $this->productRepository->existBy(['code' => $code])) {
-            throw new ProductCodeNotUniqueException($code);
+        if (!\array_key_exists('template', $data)) {
```
Prokyonn, line 57: same as above

src/Infrastructure/Doctrine/Repository/ProductFamilyRepository.php
```diff
@@ -101,8 +101,15 @@ public function createQueryBuilder(array $filters): QueryBuilder
         $productUuid = $filters['productUuid'] ?? null;
         if (null !== $productUuid) {
             Assert::string($productUuid); // @phpstan-ignore staticMethod.alreadyNarrowedType
-            $queryBuilder->innerJoin(Product::class, 'product', Join::WITH, 'product.productFamily = productFamily')
-                ->andWhere('product.uuid = :productUuid')
+            $queryBuilder
+                ->innerJoin(
+                    ProductDimensionContentInterface::class,
+                    'pdc',
```
Prokyonn: do not use abbreviations, neither as alias nor variable names

src/Infrastructure/Sulu/Admin/ProductFamilyFormMetadataVisitor.php
```diff
@@ -42,15 +42,19 @@ public function visitFormMetadata(FormMetadata $formMetadata, string $locale, ar
         $items = $formMetadata->getItems();
 
         foreach ($this->attributeGroupRepository->findAll() as $group) {
-            $groupName = $group->getTranslation($locale)?->getName() ?? '';
+            $groupTranslation = $group->getTranslation($locale)
+                ?? (($dl = $group->getDefaultLocale()) !== null ? $group->getTranslation($dl) : null);
```
Prokyonn: no abbreviations 😬

config/doctrine/Product/ProductDimensionContent.orm.xml
alexander-schranz: Think it would be `code/stage/version` if I remember unique indexes correctly NULL != NULL in Unique indexes and code is not translatable.

src/Application/MessageHandler/CreateProductMessageHandler.php
alexander-schranz: we should be consistent here to the other bundles and introduce a `ProductMapperInterface`: https://github.com/sulu/sulu/blob/4b745fc8d6a5f7413e169bfd1290e28d00b14ca9/packages/article/src/Application/MessageHandler/CreateArticleMessageHandler.php#L57-L59

Prokyonn (approved):

src/Application/MessageHandler/CreateProductMessageHandler.php
martinlagler: @Prokyonn Maybe on an import or fixture, I think most of the time there is only one product template

PR #392, opened 2026-07-07
Allow config for measurement and units

martinlagler:


src/Application/Mapper/AttributeMapper.php
```diff
@@ -31,7 +31,9 @@ public function __construct(private AttributeRepositoryInterface $attributeRepos
     public function mapAttributeData(AttributeInterface $attribute, CreateAttributeMessage|ModifyAttributeMessage $message): void
     {
         $attribute->setKey($message->getKey());
-        $attribute->setConfig($message->getConfig());
+        $config = $message->getConfig();
+        unset($config['measurementFamily']);
```
Prokyonn: is this for BC reasons? Think we can remove it

Prokyonn (approved):

PR #393, opened 2026-07-15
Add status, short description, image and documents to product main data

martinlagler:


config/doctrine/Product/ProductDimensionContent.orm.xml
```diff
@@ -36,6 +36,10 @@
 
         <field name="code" type="string" length="255" nullable="true"/>
         <field name="externalIdentifier" type="string" length="255" nullable="true" column="external_identifier"/>
+        <field name="status" type="string" length="255" nullable="true"/>
+        <field name="shortDescription" type="text" nullable="true" column="short_description"/>
+        <field name="productImage" type="integer" nullable="true" column="product_image"/>
+        <field name="productDocuments" type="json" nullable="true" column="product_documents"/>
```
alexander-schranz: I'm unsure if we should create columns here for none queried fields like (shortDescription, productImage, productDocuments) or if we maybe should introduce a `detailsData` json and save all together so it is possible to add new properties also to the product details without have the need to create new columns for things which maybe not require it.

/cc @Prokyonn @chirimoya 

```
        <field name="detailsData" type="json" nullable="true" column="product_documents"/>
```

alexander-schranz: no need for `product` prefix we are already on the product.

alexander-schranz: @chirimoya would go with a JSON field for `shortDescription, image, documents`.

@Prokyonn whats your opinion on this?

Prokyonn: yes I would also go with a JSON field, makes it easy to extend and if projects require query logic on those, they can still be extracted into a separate custom column that is kept in sync

src/Domain/Model/ProductDimensionContent.php
```diff
@@ -68,6 +68,13 @@ class ProductDimensionContent implements ProductDimensionContentInterface
      */
     protected Collection $attributes;
 
+    protected ?string $status = null;
```
Prokyonn: does it make sense for the status to be nullable? @alexander-schranz

src/Domain/Model/ProductDimensionContent.php
```diff
@@ -68,6 +68,13 @@ class ProductDimensionContent implements ProductDimensionContentInterface
...
+
+    /**
+     * @var array<string, mixed>|null
+     */
+    protected ?array $detailsData = null;
```
Prokyonn: not nullable and default empty array would make the code a bit easier with less null checks, any reason why we need it nullable?

src/Domain/Model/ProductDimensionContent.php
alexander-schranz: I would go with a default here and not make it nullable.

src/Domain/Model/ProductDimensionContent.php
alexander-schranz: Agree same as we did for templateData: https://github.com/sulu/sulu/blob/c117312b047a987d63549d82ca4500249299f37d/packages/content/src/Domain/Model/TemplateTrait.php#L23-L26

src/Application/Message/ModifyProductMessage.php
```diff
@@ -24,6 +24,10 @@
  *     template?: string,
  *     productFamily?: string,
  *     attributes?: array<int, mixed>,
+ *     status?: string|null,
+ *     shortDescription?: string|null,
+ *     image?: int|array<string, mixed>|null,
+ *     documents?: array<mixed>|null,
```
alexander-schranz: maybe also here just `details?: array<string, mixed>` to make things easier to access and also add custom props to `details` without have the need to extend something.

PR #394, opened 2026-07-15
Add frontend resolver for product detail fields

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

This adds a ProductDetailsResolver that exposes the product detail fields to the frontend under extension.product.*. The master data owned by the entity is resolved directly from the dimension content. Every field in the details bucket is resolved through the property resolver that its XML type selects, so stored media and other typed values arrive in their resolved shape. The manual media reference registration in ProductReferenceRefresher is removed because these references now flow through the same resolver pipeline.

#### Why?

The product detail fields were added in #393 but were only stored as plain data and were not available to frontend content resolution. Media fields were kept in their admin wire shape, so a resolver was needed to turn them into resolved values for the API. Registering those media references by hand was a workaround for the missing resolver and is no longer needed.

#### To Do

### DIFF
--- a/src/Infrastructure/Sulu/Content/Resolver/ProductDetailsResolver.php
+++ b/src/Infrastructure/Sulu/Content/Resolver/ProductDetailsResolver.php
@@ -41,13 +41,18 @@ class ProductDetailsResolver implements ResolverInterface
             return null;
         }
 
+        // entity master-data first; a project's `details/<field>` with a colliding bare
+        // name overrides it (array_merge — later keys win)
         return ContentView::create(
-            [
-                'code' => ContentView::create($dimensionContent->getCode(), []),
-                'externalIdentifier' => ContentView::create($dimensionContent->getExternalIdentifier(), []),
-                'productFamily' => ContentView::create($dimensionContent->getProductFamily()?->getUuid(), []),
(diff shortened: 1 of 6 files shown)


src/Domain/Model/ProductDimensionContent.php
```diff
@@ -257,4 +264,80 @@ public function removeAttribute(ProductAttributeValueInterface $attribute): stat
...
+    }
+
+    public function setStatus(?string $status): static
+    {
+        $this->status = $status;
+
+        return $this;
+    }
+
+    public function getShortDescription(): ?string
```
alexander-schranz: we could put this into a getDetailsData if we use in the `.xml` `details/shortDescription`.

So projects could add there own `details` property.

martinlagler: Should be fixed in this PR https://github.com/sulu/SuluProductBundle/pull/393

Prokyonn (approved):

PR #396, opened 2026-07-16
Add product variants

martinlagler:
#### What's in this PR?

This PR adds product variants to the bundle. Each product now carries an explicit type and a variant is a child product that references its parent. Product family attributes can be flagged as variant attributes so a variant persists only its own axis values while shared attributes stay on the parent. The admin gains a variants tab with a dedicated overlay form and publishing or unpublishing a parent cascades the transition to its variants.

#### Why?

Shops need products that share a common definition but differ along attributes such as size or color. Modeling each variant as a child product keeps the shared data on the parent and lets every variant hold only its distinguishing values. Driving the behavior from an explicit product type makes variant handling easy to find and reason about across the codebase.


src/Application/Workflow/VariantParentDirtier.php
```diff
@@ -0,0 +1,66 @@
...
+use Sulu\Content\Domain\Model\DimensionContentInterface;
+use Sulu\Content\Domain\Model\WorkflowInterface;
+use Sulu\Product\Domain\Model\ProductInterface;
+use Sulu\Product\Domain\Repository\ProductRepositoryInterface;
+
+/**
+ * Flags a variant's parent as having unpublished changes whenever the variant is created or
+ * modified — variants never publish on their own (see ApplyWorkflowTransitionProductMessageHandler).
+ */
+final class VariantParentDirtier
```
Prokyonn: @Prokyonn adjust naming and @internal

src/Infrastructure/Sulu/Admin/ProductVariantTitleBuilder.php
```diff
@@ -0,0 +1,93 @@
...
+use Sulu\Product\Domain\Model\ProductDimensionContentInterface;
+use Sulu\Product\Domain\Model\ProductFamilyAttributeInterface;
+
+/**
+ * Builds the `variantTitle` column value for the Variants tab list
+ * ({@see \Sulu\Product\UserInterface\Controller\Admin\ProductVariantController::cgetAction}).
+ *
+ * @internal
+ */
+final class ProductVariantTitleBuilder
```
Prokyonn: Should not be necessary, as variants should also get a title

src/Domain/Model/ProductInterface.php
```diff
@@ -26,10 +26,28 @@ interface ProductInterface extends AuditableInterface, ContentRichEntityInterfac
     public const FORM_KEY = 'product_details';
     public const LIST_KEY = 'products';
 
+    public const LIST_KEY_VARIANTS = 'product_variants';
+    public const FORM_KEY_VARIANT = 'product_variant';
+
+    public const TYPE_SIMPLE = 'simple';
```
Prokyonn: we will need 3 types:

```
    public const TYPE_PRODUCT = 'product';
    public const TYPE_PRODUCT_WITH_VARIANTS = 'product_with_variants';
    public const TYPE_VARIANT = 'variant';
```

still open for discussion if we choose `product`or `simple` for products without variants

src/Application/Workflow/VariantWorkflowCascader.php
```diff
@@ -0,0 +1,68 @@
...
+use Sulu\Content\Domain\Model\WorkflowInterface;
+use Sulu\Product\Domain\Model\ProductInterface;
+use Sulu\Product\Domain\Repository\ProductRepositoryInterface;
+
+/**
+ * Cascades a parent's publish/unpublish transition to each of its variants. Re-applying `publish`
+ * to an already-published variant does NOT throw (`publish` is defined `published -> published`);
+ * the catch below exists for `unpublish` on a never-published variant and missing-locale content.
+ */
+final class VariantWorkflowCascader
```
Prokyonn, line 31: if we really need this, @internal

Prokyonn (changes requested):

src/Infrastructure/Sulu/Admin/ProductAdmin.php
```diff
@@ -204,7 +204,7 @@ public function configureViews(ViewCollection $viewCollection): void
                 ->addListAdapters(['table'])
                 ->setTabTitle('sulu_product.variants')
                 ->setTabOrder(20)
-                ->setTabCondition("type == 'variant'")
+                ->setTabCondition("type == 'product_with_variants'")
```
alexander-schranz: use constant `ProductInterface::TYPE_PRODUCT_WITH_VARIANTS`

src/Domain/Model/ProductInterface.php
```diff
@@ -43,7 +44,7 @@ public function getType(): string;
 
     public function setType(string $type): self;
 
-    public function isVariantProduct(): bool;
+    public function isProductWithVariants(): bool;
```
alexander-schranz: introduce `isType(ProductInterface::TYPE_PRODUCT_WITH_VARIANTS);` instead

This let us easier search for usage of the constant to check where we do with variants specific things.

alexander-schranz (changes requested):
did not do a whole review just a small suggestion:

src/Application/Mapper/ProductParentMapper.php
```diff
@@ -0,0 +1,47 @@
...
+    ) {
+    }
+
+    public function mapProductData(ProductInterface $product, array $data): void
+    {
+        if (!$product instanceof Product) {
+            return;
+        }
+
+        if (\array_key_exists('type', $data) && \is_string($data['type'])) {
```
alexander-schranz:
```suggestion
        if (\array_key_exists('type', $data)) {
            Assert::string($data['type']);
```


Would stay in the same mechanic we use in core:

src/Domain/Model/ProductFamilyAttribute.php
```diff
@@ -19,6 +19,8 @@ class ProductFamilyAttribute implements ProductFamilyAttributeInterface
 
     protected bool $required = false;
 
+    protected bool $variant = false;
```
alexander-schranz: should we go with `variantSpecific` to be more in a domain language?

src/Infrastructure/Doctrine/Repository/ProductRepository.php
```diff
@@ -280,6 +282,17 @@ public function createQueryBuilder(array $filters, array $sortBy = [], array $se
...
+        if (null !== $parent) {
+            Assert::string($parent); // @phpstan-ignore staticMethod.alreadyNarrowedType
+            $queryBuilder->andWhere('product.parent = :parent')
+                ->setParameter('parent', $parent);
+        }
+
+        if (true === ($filters['parentIsNull'] ?? null)) {
+            $queryBuilder->andWhere('product.parent IS NULL');
+        }
+
```
alexander-schranz: Not sure about the `parentIsNull` would prefer filter by `types` or `excludeTypes`.

src/Infrastructure/Sulu/Admin/ProductAttributeFormMetadataVisitor.php
```diff
@@ -62,6 +57,9 @@ public function visitFormMetadata(FormMetadata $formMetadata, string $locale, ar
             return;
         }
 
+        $product = $this->productRepository->findOneBy(['uuid' => $id]);
+        $isProductWithVariants = null !== $product && $product->isType(ProductInterface::TYPE_PRODUCT_WITH_VARIANTS);
```
alexander-schranz: Do we need `null !== $product` can we go with something like:

```suggestion
        $isProductWithVariants = $product?->isType(ProductInterface::TYPE_PRODUCT_WITH_VARIANTS) ?? false;
```

src/Infrastructure/Sulu/Content/DataMapper/ProductAttributesDataMapper.php
```diff
@@ -111,7 +112,13 @@ public function map(
             }
         }
 
-        $this->assertRequiredSatisfied($familyAttributes, $allExisting);
+        $resource = $unlocalizedDimensionContent->getResource();
+        // On create, ProductParentMapper hasn't run yet at this point in the ContentPersister
+        // pipeline, so `isType()` alone isn't reliable — fall back to `$data['parent']`.
+        $isVariant = $resource->isType(ProductInterface::TYPE_VARIANT)
+            || (\is_string($data['parent'] ?? null) && '' !== $data['parent']);
```
alexander-schranz: Should we change priority to avoid this hack?

src/UserInterface/Controller/Admin/ProductController.php
```diff
@@ -234,7 +236,7 @@ private function getCreateData(Request $request): array
     {
         /** @var CreateProductMessageData $data */
         $data = \array_replace(
-            $request->request->all(),
+            $this->stripClientSubmittedParent($request),
```
alexander-schranz: Setting the parent sounds like something which should be a CreateVariantHandler only function and so we don't need to strip it for Modify.

alexander-schranz (changes requested):
Just some small change requests, else looks fine. Did not have a deep look at the tests.

src/UserInterface/Controller/Admin/ProductController.php
alexander-schranz: Why do we even serialize the parent here? What data do we need?

Prokyonn:
@alexander-schranz all comments addressed in https://github.com/sulu/SuluProductBundle/pull/396/changes/b6fed7781b1db1e4630cb518b2219cf72eb7f904

PR #398, opened 2026-07-24
Group product attributes into per-group sections

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Both product attribute form metadata visitors now create one section per attribute group instead of a single combined attributes section. Each section is named after the group id and labeled with the group translation for the requested locale. When that translation is missing the group default locale is used instead. If no translation resolves at all the generic attributes label is kept as a fallback.

#### Why?

Product families can define many attributes which all ended up in one flat list in the admin form. Editors already organize attributes into groups so the form should reflect that same structure. Splitting the fields into sections per group makes long attribute forms easier to scan. The variant overlay form gets the same treatment so both forms stay consistent.


alexander-schranz (approved):

alexander-schranz:
Reviewed the last 2 commits, which I think was related to this PR. Looks fine from my side.

PR #399, opened 2026-07-24
Add date attribute type and step config, remove json type

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

A new date attribute type is introduced. Date values are validated against the `Y-m-d` format and stored as UTC timestamps in the existing number column of the attribute value. The attribute form additionally exposes the step config for number attributes. The json attribute type is removed entirely together with its database column and entity accessors.

#### Why?

Product data needs date information such as release dates, which none of the existing attribute types could store reliably. Storing the timestamp in the existing number column avoids an additional schema field. The json type overlapped with the simplified scalar value model and has not been part of any release yet, so it could be removed without a BC break.

#### Example Usage

```php
$dateAttributeType = new DateAttributeType();

$dateAttributeType->writeValue($productAttributeValue, '2026-08-03'); // stored as UTC timestamp in the number column
$dateAttributeType->readValue($productAttributeValue); // returns "2026-08-03"
```

#### To Do


config/doctrine/Product/ProductAttributeValue.orm.xml
```diff
@@ -12,6 +12,7 @@
         <field name="number" type="float" nullable="true"/>
         <field name="text" type="string" length="255" nullable="true"/>
         <field name="json" type="json" nullable="true"/>
+        <field name="date" column="dateValue" type="date_immutable" nullable="true"/>
```
alexander-schranz: I would save it in `number` as unix timestamp. This make it possible that we not have issues if we add to UI range filters with >, < we not require to know where the value is stored as range is always number.

Its a common practice also in search engines handle date / datetimes as unix timestamp.

Alternative we should go with `datetime_immutable` to be flexible enough to also support datetimes in future.

src/Application/AttributeType/AbstractAttributeType.php
```diff
@@ -19,6 +19,11 @@
 
 abstract class AbstractAttributeType implements AttributeTypeInterface
 {
+    public function isAvailableInAdmin(): bool
```
alexander-schranz: Would currently not introduce it and remove it completely.

If a project require JSON it can create its own Json Type and then it need be displayed and shown in the dropdown. There are no "hidden" types.

PR #400, opened 2026-07-27
Add product associations

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Products can now reference other products through configurable association types. Each type is declared under the new `association_types` config node and receives its own field on a new associations tab in the admin. The selected targets are exposed to the frontend under `extension.associations`, and reference records and cached pages of referring products are kept in sync when a target is published or removed. The display properties of the existing product selection field type were corrected because they referenced list properties that do not exist.

#### Why?

Shops need to present related products such as alternatives or accessories on a product detail page. The product bundle had no way to model that relationship. Making the types configurable lets every project define its own vocabulary instead of relying on a fixed list shipped by the bundle.

#### Example Usage

```yaml
# config/packages/sulu_product.yaml
sulu_product:
    association_types:
        # Without a label the key sulu_product.association_type_alternative is used,
        # which the project has to provide as a translation.
        alternative: ~
        accessory:
            label: 'app.product_accessories'
```

```php
$alternatives = $productDimensionContent->getAssociationsByType('alternative');

foreach ($alternatives as $association) {
    $target = $association->getTarget();
}

// Find every product that references the given product.
$referrers = $productRepository->findBy([
    'locale' => 'en',
    'stage' => 'live',
    'associationTargetUuid' => $product->getUuid(),
]);
```

#### To Do

- [ ] Add breaking changes to UPGRADE.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,39 @@
 # Sulu Product Bundle
+
+## Association form overrides
+
+The bundle generates a `product_associations` form with one field per configured
+`sulu_product.association_types` key. A project overrides that form by shipping its own form
+XML using the same `product_associations` key in a directory registered under
+`sulu_admin.forms.directories` — the Sulu skeleton registers `config/forms` by default.
+
+Rules for the declared fields:
+
+- The field name must be `associations/<type>`, where `<type>` is a configured
+  `sulu_product.association_types` key.
+- Only the type `product_selection` is allowed — no other field type maps to product
+  associations.
+- A `properties` collection param declares which target properties resolve, in addition to the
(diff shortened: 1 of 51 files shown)


alexander-schranz (approved):

alexander-schranz:
Currently fine for me, think we need check only how we provide things to the twig as we can keep most things internal.

Issue #395, opened 2026-07-15
Product Bundle Open Tasks


martinlagler:
### Attributes

- [x] Attribute type improvements — PR #399
  - [x] Add a Date attribute type using a Unix timestamp
  - [x] Add configurable number steps / multiplier, similar to the Number property type
  - [x] Remove the JSON attribute type (removed entirely, including the value column)
- [x] Add frontend fields with placeholders, for example: > 2 GΩ — PR #401
  - Implemented as a per-attribute display format template with `%value%` and `%unit%` placeholders (e.g. `> %value% %unit%` → `> 2 Ω`)
  - The raw value stays numeric and filterable; the website receives the formatted string via the product Twig extension (`formattedValue`)
- [ ] Define a Range attribute type (open decision)
  - Option A: use two separate attributes
  - Option B: store the range as JSON (conflicts with the JSON removal in #399, would need its own storage)

### Attribute Groups

- [x] Implement attribute groups
- [x] Group product attributes into per-group sections in the product and variant forms — PR #398

### Product Family

- [x] Add a dedicated column for variant attributes — PR #396 (variant checkbox per attribute in the family form)

### Product

- [x] Add a separate section for each Product Family within the product form — covered by the per-group sections of PR #398 (a product has exactly one family; its attributes are sectioned by attribute group, the family stays visible via the Product family field)

### Variants

- [x] Product variants — PR #396
  - [x] Initially determine whether a product is a Simple Product or a Variant Product
  - [x] Add a dedicated Variants tab containing a list of all variants
  - [x] Allow variants to be created and edited using an overlay form
  - [x] Add a unique code for each variant (hidden on the main product)
  - [x] For product variants, display only attributes marked as variant attributes (overlay sections refined by #398)
- [ ] Allow variant attributes to be set on the product as a default that each variant can optionally override
- [ ] Consider automatically generating variants for all possible attribute combinations

### Website resolving
- [ ] Associations
- [ ] Attributes
- [ ] nesting

PR #401, opened 2026-08-04
Add display format for attribute values

Prokyonn:
#### What's in this PR?

Attributes of type text and number can now define an optional display format containing placeholder tokens, for example `> %value% GΩ`. The format is configured in the attribute form and stored in the existing config column. The product Twig extension applies the format and returns an additional `formattedValue` entry for every attribute, while the raw `value` stays unchanged for filtering and sorting. The `%unit%` token resolves the configured measurement unit symbol through the MeasurementRegistry and an info text in the attribute form documents the available tokens.

#### Why?

Technical product data such as an insulation resistance of `> 2 GΩ` contains qualifiers that a plain number cannot express. Storing the qualifier inside a text value would make the attribute unusable for filters. A display format keeps the stored value structured and moves the presentation concern to the website rendering.

#### Example Usage

```twig
{# attribute config: { "format": "> %value% %unit%", "unit": "OHM" }, stored value: 2 #}
{% set product = sulu_product_load(resource.id, {}) %}

{{ product.attributes[0].formattedValue }} {# > 2 Ω #}
{{ product.attributes[0].value }}          {# 2, stays numeric #}
```


config/forms/Attribute/attribute_details.xml
```diff
@@ -97,6 +97,14 @@
             </meta>
         </property>
 
+        <property name="config/format" type="text_line" colspan="6"
+                  visibleCondition='type == "text" || type == "number"'>
+            <meta>
+                <title>sulu_product.format</title>
+                <info_text>sulu_product.format_info_text</info_text>
+            </meta>
+        </property>
```
alexander-schranz: would personally name it `displayFormat` what do you think @Prokyonn ?

config/forms/Attribute/attribute_details.xml
```diff
@@ -97,6 +97,14 @@
             </meta>
         </property>
 
+        <property name="config/format" type="text_line" colspan="6"
+                  visibleCondition='type == "text" || type == "number"'>
+            <meta>
+                <title>sulu_product.format</title>
+                <info_text>sulu_product.format_info_text</info_text>
```
alexander-schranz: same

translations/admin.de.json
```diff
@@ -87,6 +87,8 @@
     "sulu_product.measurement_family_weight": "Gewicht",
     "sulu_product.min": "Min",
     "sulu_product.max": "Max",
+    "sulu_product.format": "Format",
```
alexander-schranz:
```suggestion
    "sulu_product.format": "Anzeigeformat",
```

translations/admin.en.json
```diff
@@ -86,6 +86,8 @@
     "sulu_product.measurement_family_weight": "Weight",
     "sulu_product.min": "Min",
     "sulu_product.max": "Max",
+    "sulu_product.format": "Format",
```
alexander-schranz:
```suggestion
    "sulu_product.format": "Display format",
```

alexander-schranz (changes requested):

src/Infrastructure/Symfony/Twig/ProductTwigExtension.php
```diff
@@ -117,14 +119,44 @@ private function formatAttributes(ProductDimensionContentInterface $dimensionCon
                 default => $productAttribute->getValue(),
             };
 
+            $formattedValue = match ($attribute->getType()) {
+                AttributeInterface::TYPE_TEXT => $this->formatValue($attribute, $productAttribute->getText()),
```
alexander-schranz: Confused is this the `attribute` or the `attributeValue`?

src/Infrastructure/Symfony/Twig/ProductTwigExtension.php
alexander-schranz:
```suggestion
        foreach ($dimensionContent->getAttributes() as $productAttributeValue) {
```

src/Infrastructure/Symfony/Twig/ProductTwigExtension.php
```diff
@@ -117,14 +119,44 @@ private function formatAttributes(ProductDimensionContentInterface $dimensionCon
...
+                AttributeInterface::TYPE_NUMBER => $this->formatValue($attribute, $productAttribute->getNumber()),
+                default => null,
+            };
```
alexander-schranz, line 126: What about "AttributeOption"? Do we need there also a displayFormat /cc @chirimoya

alexander-schranz (changes requested):

config/forms/Attribute/attribute_details.xml
Prokyonn: Yes I don't mind, I already used display format in the title as it feels better 😬 going to adjust it.

src/Infrastructure/Symfony/Twig/ProductTwigExtension.php
Prokyonn: Currently I don't see it, but if we see the need I would add this in a followup PR

Prokyonn:
@alexander-schranz I did also not add the resolving to the standard product resolving process yet, because there are no attributes at all. I've added the website resolving as a new open point in the https://github.com/sulu/SuluProductBundle/issues/395 issue

src/Infrastructure/Symfony/Twig/ProductTwigExtension.php
alexander-schranz: lets merge and add it maybe later. Think options can be text and numbers where a displayFormat make same sense.

alexander-schranz (approved):

PR #403, opened 2026-08-06
Fix the product versions endpoint

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | #404
| License | MIT

#### What's in this PR?

The versions tab under insights loads its list metadata but the list request itself answers 500, so the table stays empty.

* `ProductController::getVersionsAction()` passed `$listBuilder->getCurrentPage()` straight into `PaginatedRepresentation`. That getter returns the raw query value as a string while the constructor argument is `?int`, and this file declares `strict_types`, so the call throws. It is now cast, which is what every other list action in the bundle already does. The same uncast call sits in the core page, article and snippet controllers, but those files do not declare `strict_types`, so PHP coerces the value and they get away with it.
* The `products_versions` literal appeared in three places, the resource registration in `SuluProductBundle` and both the field descriptor lookup and the representation rel in the controller. It is now `ProductInterface::LIST_KEY_VERSIONS`, defined once.

Tests:

* `ProductVersionsViewTest` covers the two things the tab needs before it ever reaches the endpoint, that the view's `listKey` resolves to real list metadata and that its `resourceKey` is registered with a list route. The content bundle derives the versions keys from the dimension content resource key, so the registered `products_versions` list is only found while that key stays aligned with the product one. Splitting them again would break the tab silently, and these fail instead.
* `ProductControllerTest::testGetVersionsListsPublishedVersions` calls the endpoint with the exact query the tab sends, and asserts the paginated envelope the table reads.

Verified in the browser. The table renders both versions of a product with title, publish date and changer, and restoring a version returns 200 and navigates back to the details view.

#### Why?

The tab has never worked. It failed twice over, first with a 404 on `/admin/metadata/list/product_contents_versions` and then with this 500. #404 dropped the separate dimension content resource key, which fixed the first failure and made the workaround this PR originally carried unnecessary, so only the endpoint fix and its tests remain.

#### To Do

### DIFF
--- a/src/Domain/Model/ProductInterface.php
+++ b/src/Domain/Model/ProductInterface.php
@@ -29,6 +29,8 @@ interface ProductInterface extends AuditableInterface, ContentRichEntityInterfac
     public const LIST_KEY_VARIANTS = 'product_variants';
     public const FORM_KEY_VARIANT = 'product_variant';
 
+    public const LIST_KEY_VERSIONS = 'products_versions';
+
     public const TYPE_PRODUCT = 'product';
     public const TYPE_PRODUCT_WITH_VARIANTS = 'product_with_variants';
     public const TYPE_VARIANT = 'variant';
--- a/src/Infrastructure/Symfony/HttpKernel/SuluProductBundle.php
+++ b/src/Infrastructure/Symfony/HttpKernel/SuluProductBundle.php
@@ -1215,7 +1215,7 @@ final class SuluProductBundle extends AbstractBundle
                                 'detail' => 'sulu_product.get_product',
                             ],
                         ],
-                        'products_versions' => [
+                        ProductInterface::LIST_KEY_VERSIONS => [
                             'routes' => [
                                 'list' => 'sulu_product.get_product_versions',
                                 'detail' => 'sulu_product.get_product',
--- a/src/UserInterface/Controller/Admin/ProductController.php
+++ b/src/UserInterface/Controller/Admin/ProductController.php
@@ -223,7 +223,7 @@ final class ProductController implements SecuredControllerInterface
         $locale = $this->getLocale($request);
 
         /** @var DoctrineFieldDescriptorInterface[] $fieldDescriptors */
-        $fieldDescriptors = $this->fieldDescriptorFactory->getFieldDescriptors('products_versions');
+        $fieldDescriptors = $this->fieldDescriptorFactory->getFieldDescriptors(ProductInterface::LIST_KEY_VERSIONS);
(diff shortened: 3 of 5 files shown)


src/Infrastructure/Sulu/Admin/ProductContentAdmin.php
```diff
@@ -54,6 +57,16 @@ public function configureViews(ViewCollection $viewCollection): void
...
+
+            // the content bundle derives the versions keys from the dimension content resource
+            // key, which is "product_contents" here, but the list is registered as
+            // "products_versions"
+            if (self::VERSIONS_VIEW === $viewBuilder->getName() && $viewBuilder instanceof ListViewBuilderInterface) {
+                $viewBuilder
+                    ->setResourceKey(ProductInterface::LIST_KEY_VERSIONS)
+                    ->setListKey(ProductInterface::LIST_KEY_VERSIONS);
+            }
+
```
alexander-schranz: but is this a bug? How this this not fail on article bundle?

Prokyonn: Article can't hit it because it uses the same key everywhere: `ArticleInterface::RESOURCE_KEY`, `ArticleDimensionContent::getResourceKey()`, the preview provider tag and the route defaults provider are all `articles`. `createViews()` derives the versions key from `$dimensionContentClass::getResourceKey()`, so it lands on `articles_versions`, which is exactly what `articles_versions.xml` declares. Pages and snippets the same.

We have two keys instead of one. `ProductInterface::RESOURCE_KEY` is `products`, but the dimension content says `product_contents` (that's also the preview provider key and the reference resource key). So the derivation gives us `product_contents_versions`, which nothing declares. The admin requests `/admin/metadata/list/product_contents_versions`, gets a 404, and the tab body stays empty.

Prokyonn: If we want the override gone entirely, the move is to have `ProductDimensionContent::getResourceKey()`
return `products` like core does, then the derivation just works on its own. It drags a few things
along though: the preview provider tag, the route defaults provider, the reference resource key and
the `product_contents` resource entry all point at that string today.

alexander-schranz: Sounds good to go the same way as the article bundle now as from our perspective its now one resources and not content an own resource like in the previous implementation.

alexander-schranz (approved):

PR #406, opened 2026-08-10
Fix the product list publish indicator and icon

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

* **The publish indicator in the product list never showed a state.** `AbstractTableAdapter` reads `const draft = !item.publishedState` and `const published = !!item.published`. The list returned the raw workflow place as a string, which is always truthy, so `draft` was permanently false, and no `published` property was returned at all. Every row rendered the same, whether it was published, unpublished or carrying draft changes.
* `config/lists/products.xml` gains the `published` property that the articles list already declares, mapped to `workflowPublished` with a datetime transformer.
* `ProductController::cgetAction()` selects `published` and `publishedState` whether or not the list asked for them, then maps the workflow place to a boolean. This mirrors what `ArticleController` does.
* **The navigation icon changes from `su-newspaper` to `su-industry`.** The selection overlays and the search result entry move with it, so a product looks the same wherever it appears. `ProductLinkProvider` keeps `su-document`, which is what the article link provider uses too.

Before and after for the same list request, with neither field requested:

```
{"publishedState": "draft"}
{"publishedState": false, "published": "2026-07-23T18:04:31+00:00"}
```

#### Why?

The indicator is the only place in the list that tells an editor whether a product is live, so a product with unpublished changes looked exactly like one that was fully published.

The new functional test covers both states, including the unpublished one, and asserts that the two fields come back even when the request asks only for `name` and `id`.

#### To Do

### DIFF
--- a/config/lists/products.xml
+++ b/config/lists/products.xml
@@ -180,6 +180,15 @@
             <entity-name>dimensionContent</entity-name>
         </property>
 
+        <property name="published" translation="sulu_admin.published" type="date" visibility="never">
+            <field-name>workflowPublished</field-name>
+            <entity-name>dimensionContent</entity-name>
+
+            <joins ref="dimensionContent"/>
+
+            <transformer type="datetime"/>
+        </property>
+
         <property name="publishedState" translation="sulu_content.published_state" visibility="never">
             <field-name>workflowPlace</field-name>
             <entity-name>dimensionContent</entity-name>
--- a/src/Infrastructure/Sulu/Admin/ProductAdmin.php
+++ b/src/Infrastructure/Sulu/Admin/ProductAdmin.php
@@ -61,7 +61,7 @@ class ProductAdmin extends Admin
 
         $parent = new NavigationItem('sulu_product.products');
         $parent->setPosition(20);
-        $parent->setIcon('su-newspaper');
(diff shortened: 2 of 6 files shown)


alexander-schranz:
I thought this was expected refactoring that we not longer build in 3.0 on boolean states on publishState and go with `draft`, ... didn't we already update that in some JS components? I would prefer not including again a state which build ontop of publishedState and published date. Instead refactoring the JS components to work with a single state.

Prokyonn:
> I thought this was expected refactoring that we not longer build in 3.0 on boolean states on publishState and go with `draft`, ... didn't we already update that in some JS components? I would prefer not including again a state which build ontop of publishedState and published date. Instead refactoring the JS components to work with a single state.

Unfortunately we never did refactor this properly, articles are also using publishState as boolean:
https://github.com/sulu/sulu/blob/3.0/packages/article/src/UserInterface/Controller/Admin/ArticleController.php#L132

alexander-schranz (approved):

PR #407, opened 2026-08-24
Update product resolving

martinlagler:


src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -0,0 +1,300 @@
...
+use Sulu\Product\Domain\Model\ProductDimensionContentInterface;
+use Sulu\Product\Domain\Model\ProductInterface;
+use Sulu\Product\Infrastructure\Sulu\Content\ResourceLoader\ProductFamilyResourceLoader;
+use Sulu\Product\Infrastructure\Sulu\Content\ResourceLoader\ProductResourceLoader;
+
+/**
+ * Assembles the root-level `product` namespace: the Details master-data fields flattened onto it,
+ * plus `attributes`, `associations` and `variants`.
+ *
+ * How much resolves depends on why: a non-null `$properties` means this product is a reference on
```
Prokyonn: make this comment more concise or remove it

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -0,0 +1,300 @@
...
+            return null;
+        }
+
+        // merged content always carries a locale
+        /** @var string $locale */
+        $locale = $dimensionContent->getLocale();
+
+        $content = $this->resolveDetails($dimensionContent, $locale);
+
+        if (null === $properties) {
```
Prokyonn: Why do we resolve attributes only when properties === null?

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -0,0 +1,300 @@
...
+            $content['attributes'] = ContentView::create($this->resolveAttributes($dimensionContent, $locale), []);
+        }
+
+        $content['associations'] = ContentView::create($this->resolveAssociations($dimensionContent, $locale), []);
+
+        if (null === $properties) {
+            $variants = $this->resolveVariants($dimensionContent, $locale);
```
Prokyonn, line 87: same here

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -0,0 +1,300 @@
...
+            $formatted = $this->formatValue($value, $attribute, $locale);
+
+            if (null === $formatted || '' === $formatted) {
+                continue;
+            }
+
+            $group = $attribute->getGroup();
+            $groupId = (string) $group->getId();
+
+            $attributes[$attribute->getKey()] = [
```
Prokyonn: Why do we not use the ProductAttributeValue directly?

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -0,0 +1,300 @@
...
+                ?? $value->getAttributeOptionKey(),
+            AttributeInterface::TYPE_TEXT => $this->applyDisplayFormat($attribute, $value->getText()),
+            AttributeInterface::TYPE_NUMBER => $this->applyDisplayFormat($attribute, $value->getNumber()),
+            AttributeInterface::TYPE_DATE => $this->formatDate($value->getNumber(), $locale),
+            default => null,
+        };
+    }
+
+    /** An editor can set a unit and a display format per attribute; without them the value renders bare. */
+    private function applyDisplayFormat(AttributeInterface $attribute, string|float|null $value): ?string
```
Prokyonn: This is already part of the twig extension, so why duplicate it here? I think we could directly use the ProductAttributeValue and the rest can be done from the rendering side

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -0,0 +1,300 @@
...
+    /**
+     * The association targets under `product.associations.*`. Which properties resolve per type is
+     * controlled by the merged `product_associations` form metadata — projects override it by
+     * declaring `associations/<type>` fields with params.
+     *
+     * @return array<string, mixed>
+     */
+    private function resolveAssociations(ProductDimensionContentInterface $dimensionContent, string $locale): array
+    {
+        // by design a missing warmed form cache fails loudly here (MetadataNotFoundException)
```
Prokyonn: comment can be removed

src/Infrastructure/Sulu/Content/ProductFamilyWrapper.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+namespace Sulu\Product\Infrastructure\Sulu\Content;
+
+use Sulu\Product\Domain\Model\ProductFamilyInterface;
+
+/**
+ * A family as a template sees it: one locale, no entity graph.
+ */
+final readonly class ProductFamilyWrapper
```
Prokyonn: I dont see a reason for this class

src/Infrastructure/Sulu/Content/ResourceLoader/ProductFamilyResourceLoader.php
```diff
@@ -0,0 +1,60 @@
...
+use Sulu\Content\Application\ResourceLoader\Loader\ResourceLoaderInterface;
+use Sulu\Product\Domain\Repository\ProductFamilyRepositoryInterface;
+use Sulu\Product\Infrastructure\Sulu\Content\ProductFamilyWrapper;
+
+/**
+ * @internal if you need to override this service, create a new service with based on ResourceLoaderInterface instead of extending this class
+ *
+ * @final
+ */
+class ProductFamilyResourceLoader implements ResourceLoaderInterface
```
Prokyonn: See my comment above, I think we dont need this

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -0,0 +1,300 @@
...
+
+    private function resolveProductFamily(ProductDimensionContentInterface $dimensionContent): ContentView
+    {
+        $uuid = $dimensionContent->getProductFamily()?->getUuid();
+
+        if (null === $uuid) {
+            return ContentView::create(null, []);
+        }
+
+        return ContentView::createResolvable(
```
Prokyonn: This does not make sense, the ProductFamily should already be eager loaded here, so why use a ResourceLoader to do another query? I think we should just directly use the ProductFamily entity here?

src/Application/Variant/ProductVariantLoader.php
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Loads a product's variants as merged dimension content.
+ *
+ * Variants are not routable, so a project addresses them by code rather than by URL.
+ *
+ * @internal
+ *
+ * @final
+ */
+readonly class ProductVariantLoader
```
Prokyonn: Do we have multiple use cases for this class? Otherwise I don't see a reason yet for extracting this 🤔

UPGRADE.md
```diff
@@ -0,0 +1,84 @@
+# Upgrade
+
+## Product data moved to a root-level `product` namespace
```
Prokyonn: This bundle is in heavy development. We do not require an upgrade.md before the first release

src/Infrastructure/Symfony/Twig/ProductTwigExtension.php
```diff
@@ -49,7 +46,7 @@ public function getFunctions(): array
     /**
      * @param array<string, string> $properties
      *
-     * @return array{attributes: list<array{key: string, label: string, type: string, value: mixed, formattedValue: string|null}>, ...}|null
+     * @return array<string, mixed>|null
      */
     public function loadProduct(
```
Prokyonn, line 51: `$properties` is a required, non-nullable `array` here, so a template can never pass `null`.
The resolver only builds `attributes` and `variants` when `properties === null`, so
`sulu_product_load()` can never return either of them anymore.

Prokyonn (changes requested):

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
martinlagler: This is for smart_content and selections so if specific properties are given, we may not want to resolve all variants and attributes... But i can change this.

I think this comment here also caused this 
https://github.com/sulu/sulu/pull/9046/changes#diff-c698cbb1f2afce655c27e32ebfc8d9f37af04b8164a1d8d609da71d2858c6697R40-R42

Prokyonn: yes but now we cannot even set specific properties or am I missing something?

Prokyonn (approved):

PR #409, opened 2026-08-27
Optimize queries

martinlagler:


src/Domain/Repository/AttributeRepositoryInterface.php
```diff
@@ -44,6 +45,11 @@ public function getOneBy(array $filters): AttributeInterface;
     /** @param array<string, mixed> $criteria */
     public function countBy(array $criteria): int;
 
+    /**
+     * @return list<AttributeInterface>
+     */
+    public function findByFamily(ProductFamilyInterface $family): array;
```
alexander-schranz: Just stumbled over this `findBy(['familyId' => $familyId])` we should go over the repositories in general they seems not match our patterns. e.g. `findAll` is in our cases `findBy` without filter and so on.

src/Infrastructure/Doctrine/Repository/AttributeGroupRepository.php
```diff
@@ -101,8 +101,22 @@ public function createQueryBuilder(array $filters): QueryBuilder
 
     public function findAll(): array
```
Prokyonn: should be replaced by findBy

src/Infrastructure/Doctrine/Repository/ProductFamilyRepository.php
```diff
@@ -118,9 +138,70 @@ public function createQueryBuilder(array $filters): QueryBuilder
...
+     * @param ProductFamilyRepositorySelects $selects
+     */
+    private function addRowMultiplyingSelects(ProductFamilyInterface $family, array $selects): void
+    {
+        if (!($selects[self::SELECT_FAMILY_ATTRIBUTE_OPTIONS] ?? false)) {
+            return;
+        }
+
+        // Hydrating the options fills the attributes' collections, the result itself is not needed.
+        $this->entityManager->createQueryBuilder()
```
Prokyonn: Not really sure about this additional query 🤔  but I guess we do not really have other optimized solutions?

src/Infrastructure/Sulu/Content/Normalizer/ProductAttributesNormalizer.php
```diff
@@ -51,8 +53,8 @@ public function enhance(object $object, array $normalizedData): array
 
         $productFamily = $object->getProductFamily();
         if (null !== $productFamily) {
-            foreach ($productFamily->getFamilyAttributes() as $familyAttribute) {
-                $attribute = $familyAttribute->getAttribute();
+            // One query for the whole family, walking the family attributes costs one per attribute.
```
Prokyonn: shouldnt the family already be eager loaded here?

Prokyonn (changes requested):

src/Infrastructure/Doctrine/Repository/ProductFamilyRepository.php
martinlagler: Hmm couldn't find any good alternative.

config/doctrine/Product/ProductDimensionContent.orm.xml
```diff
@@ -40,7 +40,8 @@
         <field name="detailsData" type="json" nullable="false" column="details_data"/>
 
         <many-to-one field="productFamily"
```
Prokyonn, line 42: we dont set eager here, we build the sql query manually in the repository via the createQueryBuilder function and when we select / join it, it will already be loaded, so we do not need additional queries

The querybuilder method is in control of what is fetched and what is not fetched, depending on what is required.

You can check the existing Page/Article repositories

src/Infrastructure/Doctrine/Repository/ProductFamilyRepository.php
```diff
@@ -118,9 +138,70 @@ public function createQueryBuilder(array $filters): QueryBuilder
...
 
+        // selects
+        if ($selects[self::SELECT_FAMILY_ATTRIBUTES] ?? false) {
+            // Translations stay unfiltered, callers fall back to the default locale.
+            $queryBuilder
+                ->addSelect('familyAttribute', 'attribute', 'attributeGroup', 'attributeTranslation')
+                ->leftJoin('productFamily.familyAttributes', 'familyAttribute')
+                ->leftJoin('familyAttribute.attribute', 'attribute')
+                ->leftJoin('attribute.group', 'attributeGroup')
+                ->leftJoin('attribute.translations', 'attributeTranslation');
```
Prokyonn, line 167: This still leaves one gap for the two form visitors: they call `$group->getTranslation($locale)` in `createGroupSection()`, and `AttributeGroup::getTranslation()` goes through `Criteria::matching()`. On an uninitialised collection that runs a query on every call and never initialises the collection, so each group still costs 1-2 queries (locale + default-locale fallback).

I measured it with a throwaway test counting `Com_select`: attribute + option walk = 0 queries, group translation walk = 2 queries for a single group.

Suggest hydrating `attributeGroup.translations` in `addRowMultiplyingSelects()` as well (either a third small query, or via the group in the options query).

tests/Functional/Repository/ProductFamilyRepositoryTest.php
```diff
@@ -158,4 +177,79 @@ public function testRemoveDeletesFromDatabase(): void
...
+        $this->entityManager->clear();
+
+        $family = $this->repository->findOneBy(
+            ['uuid' => $uuid],
+            [ProductFamilyRepositoryInterface::GROUP_SELECT_PRODUCT_FAMILY_FORM => true],
+        );
+        $this->assertInstanceOf(ProductFamilyInterface::class, $family);
+
+        // family, family attribute, attribute, group, attribute translation, option, option translation
+        $this->assertSame(7, $this->entityManager->getUnitOfWork()->size());
```
Prokyonn: `getUnitOfWork()->size()` can't catch the gap above: the fixture has no group translation, so 7 is correct while the collection is still missing. I'd switch these to the query-count pattern from `ProductRepositoryTest` and walk the graph the way the visitors do (`getTranslation()` on attribute, option and group), with a fixture that has a translation on each level.

tests/Functional/Repository/AttributeGroupRepositoryTest.php
```diff
@@ -74,17 +74,39 @@ public function testCreateGeneratesUniqueUuids(): void
         $this->assertNotSame($a->getUuid(), $b->getUuid());
     }
 
-    public function testFindAllReturnsAllGroups(): void
+    public function testFindByWithoutFiltersReturnsAllGroups(): void
```
Prokyonn, line 78: Nothing here fails if the fetch joins (or the second query) in `findBy()` are removed; both tests only count and filter. Same ask as for the family repository: a query-count assertion over a walk through `getTranslation()` on the group and on its attributes. (I checked that the two-query technique does work: walk = 0 queries.)

src/Infrastructure/Doctrine/Repository/ProductRepository.php
```diff
@@ -407,6 +409,14 @@ public function createQueryBuilder(array $filters, array $sortBy = [], array $se
                 $dimensionAttributes,
                 $contentSelects,
             );
+
+            if ($selects[self::SELECT_PRODUCT_FAMILY] ?? false) {
```
Prokyonn: `with-product-family` silently does nothing without `with-product-content`, because the join hangs off `dimensionContent`. The interface docblock lists it as an independent key. Sulu core has the same dependency for `SELECT_EXCERPT_CATEGORIES_TRANSLATION` and guards it with `Assert::notFalse(...)` (`DimensionContentQueryEnhancer::addSelects()`); let's do the same here.

src/Domain/Repository/ProductFamilyRepositoryInterface.php
```diff
@@ -22,22 +22,43 @@
...
+    public const SELECT_FAMILY_ATTRIBUTE_OPTIONS = 'with-family-attribute-options';
+
     public function create(): ProductFamilyInterface;
 
     /**
      * @param ProductFamilyRepositoryFilters $filters
+     * @param ProductFamilyRepositorySelects $selects
      */
-    public function findOneBy(array $filters): ?ProductFamilyInterface;
+    public function findOneBy(array $filters, array $selects = []): ?ProductFamilyInterface;
```
Prokyonn, line 54: Heads-up for the rebase: #407 added `findBy(array $filters = []): iterable` to this interface. After rebasing, `findOneBy`/`getOneBy` take `$selects` and `findBy` doesn't; please give `findBy` the same parameter (no limit there, so the to-many join is safe).

src/Infrastructure/Doctrine/Repository/AttributeGroupRepository.php
```diff
@@ -96,13 +98,31 @@ public function createQueryBuilder(array $filters): QueryBuilder
                 ->setParameter('externalIdentifier', $externalIdentifier);
         }
 
+        foreach ($sortBy as $field => $direction) {
+            $queryBuilder->addOrderBy('attributeGroup.' . $field, $direction);
```
Prokyonn: This interpolates whatever key comes in into the DQL. `ProductRepository` and the core `PageRepository` whitelist the sort fields by name; let's match that. Alternatively drop `$sortBy` for now: neither caller passes it, and without an `ORDER BY` the sections in the family form come out in DB order.

src/Infrastructure/Doctrine/Repository/ProductFamilyRepository.php
```diff
@@ -50,9 +61,10 @@ public function create(): ProductFamilyInterface
         return $family;
     }
 
-    public function findOneBy(array $filters): ?ProductFamilyInterface
+    public function findOneBy(array $filters, array $selects = []): ?ProductFamilyInterface
     {
-        $queryBuilder = $this->createQueryBuilder($filters);
+        $selects = $this->resolveSelectGroups($selects);
```
Prokyonn: `resolveSelectGroups()` runs here and again inside `createQueryBuilder()`. `ProductRepository` resolves once, in `createQueryBuilder()`. Could `addRowMultiplyingSelects()` resolve for itself instead, so each entry point does it once?

src/Infrastructure/Sulu/Content/Normalizer/ProductAttributesNormalizer.php
```diff
@@ -51,6 +51,7 @@ public function enhance(object $object, array $normalizedData): array
 
         $productFamily = $object->getProductFamily();
         if (null !== $productFamily) {
+            // SELECT_PRODUCT_FAMILY joins the family graph onto the product, so this walk costs no query.
```
Prokyonn: This comment states something that is decided in another file and is only conditionally true (see the `with-product-content` thread). I'd drop it; the normalizer can't guarantee it.

src/Infrastructure/Symfony/Serializer/Normalizer/ProductFamilyNormalizer.php
```diff
@@ -45,7 +45,7 @@ public function normalize(mixed $data, ?string $format = null, array $context =
         }
 
         $attributes = [];
-        foreach ($this->attributeGroupRepository->findAll() as $group) {
+        foreach ($this->attributeGroupRepository->findBy() as $group) {
```
Prokyonn: Minor: this caller only reads `getId()`, so the second query in `findBy()` (group translations) is a wasted round trip here. Fine as is, just noting that `findBy()` eager-loads unconditionally while the other repositories gate joins behind selects.

src/Domain/Repository/ProductFamilyRepositoryInterface.php
```diff
@@ -22,22 +22,43 @@
...
+ * @phpstan-type ProductFamilyRepositorySelects array{
+ *     product_family_form?: bool,
+ *     with-family-attributes?: bool,
+ *     with-family-attribute-options?: bool,
+ * }|array<string, mixed>
  */
 interface ProductFamilyRepositoryInterface
 {
+    /**
+     * Groups are used in controllers and represents serialization / resolver group,
```
Prokyonn: Nit: "used in controllers" is copied from `ProductRepositoryInterface`; this group is used by the form metadata visitors.

Prokyonn (approved):

PR #413, opened 2026-08-31
Fix variants order + variants urls

martinlagler:


src/Infrastructure/Sulu/Content/ResourceLoader/ProductResourceLoader.php
```diff
@@ -55,11 +74,104 @@ public function load(array $ids, ?string $locale, array $params = []): array
...
+                }
+            }
+        }
+
+        if ([] === $uuids) {
+            return;
+        }
+
+        /** @var list<array{uuid: string, slug: string|null}> $rows */
+        $rows = $this->entityManager->createQueryBuilder()
```
Prokyonn: why did you not use the ProductRepository here?

Prokyonn: Seconding this, with a test consequence: because the query is built here, `ProductResourceLoaderTest` has to stub the whole query builder (`select`/`from`/`join`/`leftJoin`/`where`/`andWhere`/`setParameter` on `Argument::cetera()`), so nothing verifies that `dimensionContent.route` and `dimensionContent.product` are real associations, that `route.slug` exists, or that `getArrayResult()` returns the `{uuid, slug}` shape the code destructures. I checked those by hand and they hold (`route` is mapped by `Sulu\Content\Infrastructure\Doctrine\MetadataLoader:118` for `RoutableInterface`), but a regression would stay green. Behind a repository method the DQL would be covered by the repository's own functional tests and the mock disappears.

src/Infrastructure/Sulu/Content/ResourceLoader/ProductResourceLoader.php
```diff
@@ -55,11 +74,104 @@ public function load(array $ids, ?string $locale, array $params = []): array
...
+        if (!$product->isType(ProductInterface::TYPE_VARIANT)
+            || null === $parent
+            || null === $code
+            || '' === $code
+            || null === $locale
+        ) {
+            return ContentView::create([], []);
+        }
+
+        $slug = $this->parentSlugs[$parent->getUuid() . '|' . $locale] ?? null;
```
Prokyonn: This lookup keys on `$resource->getLocale()`, but `collectParentSlugs()` keyed on the locale that was passed to `load()`. Those are the same locale only until a shadow kicks in: `ContentResolver.php:203-207` re-aggregates the child at `$context['_shadowLocale']` when it has no template key in the page locale, so the dimension content that arrives here carries the shadow locale and the map only holds `<parentUuid>|<pageLocale>`. The lookup misses, `$slug` is null, and the variant URL silently disappears on shadow pages. Reading the request locale from somewhere the map agrees with, or keying the map by parent uuid alone within a single-locale `load()`, would close it.

src/Infrastructure/Doctrine/Repository/ProductRepository.php
```diff
@@ -246,12 +262,13 @@ public function removeDimensionContent(DimensionContentInterface $dimensionConte
      *     page?: int,
      *     limit?: int,
      * } $filters
-     * @param array{
-     *     uuid?: 'asc'|'desc',
-     *     title?: 'asc'|'desc',
-     *     created?: 'asc'|'desc',
-     *     position?: 'asc'|'desc',
-     * } $sortBy
+     * @param array<string, 'asc'|'desc'> $sortBy keyed by any field in self::SORT_ALIASES, in the
```
Prokyonn: The shape belongs on the interface, and there it still says something else. `ProductRepositoryInterface.php:110` and `:145` declare `array{id?: 'asc'|'desc', title?: 'asc'|'desc'}` for `findBy()` and `findIdentifiersBy()`, so the contract promises `id` (not in `SORT_ALIASES`, and products key on `uuid`) and omits `position`, which is exactly what `ProductResolver::findVariantUuids()` passes. A caller typed against the interface cannot use this fix without a PHPStan error. Worth moving the new shape up and dropping `id`.

tests/Functional/Repository/ProductRepositoryTest.php
```diff
@@ -622,6 +622,48 @@ public function testCreateQueryBuilderWithUuidFilterReturnsExpectedResult(): voi
...
+    public function testCreateQueryBuilderSortsInTheOrderTheKeysWereGiven(): void
+    {
+        $qb = $this->doctrineRepository->createQueryBuilder(
+            ['locale' => 'en', 'stage' => 'live'],
+            ['position' => 'asc', 'created' => 'asc', 'uuid' => 'asc'],
+        );
+
+        $this->assertStringEndsWith(
+            'ORDER BY product.position asc, product.created asc, product.uuid asc',
+            $qb->getDQL(),
```
Prokyonn, line 639: These three tests stop at `getDQL()`, which only concatenates the DQL parts. Parsing against the metadata happens in `getQuery()`, so a wrong alias or an unmapped field in `SORT_ALIASES` passes green here. That matters for the three new entries: `changed`, `authored` and `workflowPublished` have no caller in `src/`, so nothing in the suite executes them at all. One `->getQuery()->getResult()`, or an assertion on the actual row order for three products with different `position` and `created`, would cover both the clause order and the mapping. The file is already functional, so it is cheap.

src/Infrastructure/Doctrine/Repository/ProductRepository.php
```diff
@@ -374,16 +394,15 @@ public function createQueryBuilder(array $filters, array $sortBy = [], array $se
...
-                } elseif ('created' === $field) {
-                    $queryBuilder->addOrderBy('product.created', $order);
-                } elseif ('position' === $field) {
-                    $queryBuilder->addOrderBy('product.position', $order);
-                }
+        // One pass over the caller's keys, so their order is the ORDER BY order.
+        foreach ($sortBy as $field => $order) {
+            $alias = self::SORT_ALIASES[$field] ?? null;
+
+            if (null === $alias || ('filterDimensionContent' === $alias && !$hasDimensionContentJoin)) {
```
Prokyonn: An unknown key being ignored is right, but a known one is a different case: `['title' => 'asc']` without `locale`/`stage` is a caller mistake that yields arbitrarily ordered rows and no signal. #409 took the opposite line for the same class of problem in this method, where `with-product-family` without `with-product-content` throws (`Assert::notFalse`, line 415). Either the same treatment here, or split the two branches so the silent one is only the unknown field.

src/Infrastructure/Sulu/Content/ResourceLoader/ProductResourceLoader.php
```diff
@@ -55,11 +74,104 @@ public function load(array $ids, ?string $locale, array $params = []): array
...
         return self::RESOURCE_LOADER_KEY;
     }
+
+    public function reset(): void
+    {
+        $this->parentSlugs = [];
+    }
+
+    /**
+     * One query for every parent in the batch, so a listing of variants costs the same as one.
```
Prokyonn: "One query for every parent in the batch" reads as one query per parent, which is the opposite of what it does. Something like "One query for the whole batch, so a listing of variants costs the same as one".

src/Infrastructure/Symfony/HttpKernel/SuluProductBundle.php
```diff
@@ -227,6 +227,11 @@ public function configure(DefinitionConfigurator $definition): void
                         ->end()
                     ->end()
                 ->end()
+                ->scalarNode('variant_query_parameter')
```
Prokyonn, line 232: Nothing in the bundle reads `?variant=<code>` back off the request, so this key is a contract between the URL the loader writes and the parsing a project has to do itself. Right now the only place that is stated is the `->info()` string. The README documents `association_types`; this deserves a line there too, since a project that changes the key has to match it by hand on the read side.

Prokyonn (approved):

PR #414, opened 2026-09-01
Refactoring product family attributes styling

martinlagler:
Related to this PR: https://github.com/sulu/sulu/pull/9063

Beyond the titled change this drops the `sulu/sulu` fork repository from `composer.json` and requires `^3.1@dev` instead. Two adaptions come with it: `ProductResolver` implements the new `getType()`/`getOutputPath()` of `ResolverInterface`, and the order-by assertions in `ProductRepositoryTest` normalise the direction casing, because doctrine/orm 3.7 renders `ASC`/`DESC` upper case.


src/Domain/Exception/InvalidProductFamilyAttributesException.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+namespace Sulu\Product\Domain\Exception;
+
+/**
+ * Thrown when a submitted "attributes" list is non-empty but no entry parses into
+ * {id: string, required?: bool, variantSpecific?: bool} — the shape an old client build sends looks
+ * exactly like "remove every attribute", which must not happen silently.
+ */
+class InvalidProductFamilyAttributesException extends \Exception
```
Prokyonn: Not sure if we really need this exception, is this only due to the reason this was changed in this PR? Because we won't have any "old clients" yet as this bundle is unreleased

Prokyonn: This exception can go, together with the 422 handling in `postAction`/`putAction` and the two 422 tests. The bundle has no 3.x release, so there is no old client that could send the previous shape, and `extractAttributes()` dropping entries it cannot parse is the behaviour the endpoint had before.

src/Infrastructure/Symfony/HttpKernel/SuluProductBundle.php
```diff
@@ -822,14 +818,6 @@ public function loadExtension(array $config, ContainerConfigurator $container, C
...
 
-        $services->set('sulu_product.product_family_form_metadata_visitor')
-            ->class(ProductFamilyFormMetadataVisitor::class)
-            ->args([
-                new Reference('sulu_product.attribute_group_repository'),
-                new Reference('translator'),
-            ])
-            ->tag('sulu_admin.form_metadata_visitor');
-
         $services->set('sulu_product.attribute_field_factory')
```
Prokyonn, line 821: The service registration is gone here, but `src/Infrastructure/Sulu/Admin/ProductFamilyFormMetadataVisitor.php` and its unit test are still in the tree, 285 lines that can no longer run and that still render the `sulu_product.attribute_enabled` key this PR deletes. Can both files go with the rest of the old form?

src/Application/Mapper/ProductFamilyMapper.php
```diff
@@ -60,55 +59,38 @@ private function mapAttributes(
...
 
         $existingMap = [];
         foreach ($family->getFamilyAttributes() as $familyAttribute) {
-            $existingMap[$familyAttribute->getAttribute()->getId()] = $familyAttribute;
-        }
-
-        foreach ($existingMap as $attributeId => $familyAttribute) {
-            if (!isset($enabledAttributes[$attributeId])) {
+            $uuid = $familyAttribute->getAttribute()->getUuid();
+            if (null === $uuid || !isset($submitted[$uuid])) {
```
Prokyonn: The null check is needed for the `?string` return type, but letting it fall into `removeFamilyAttribute()` means a missing uuid silently deletes the assignment. A `continue` (or an assert) says more accurately that this state should not occur.

src/UserInterface/Controller/Admin/AttributeController.php
```diff
@@ -69,7 +69,14 @@ public function cgetAction(Request $request): Response
         $listBuilder = $this->listBuilderFactory->create(AttributeInterface::class);
         $listBuilder->setIdField($fieldDescriptors['id']);
         $this->restHelper->initializeListBuilder($listBuilder, $fieldDescriptors);
+
+        foreach (['group', 'groupName', 'position'] as $fieldName) {
+            $listBuilder->addSelectField($fieldDescriptors[$fieldName]);
+        }
+
+        // Positions count per group, so the default order is group name first.
```
Prokyonn, line 79: The comment covers the sort but not the three forced select fields, and their reason lives in core: `setSelectFields()` skips `visibility="never"` fields, and `ProductFamilyAttributes.js` needs group, groupName and position. Naming `product_family_attributes` as the consumer here saves the next reader the cross repo hop.

composer.json
```diff
@@ -21,7 +13,7 @@
         "doctrine/orm": "^2.17.3 || ^3.3",
         "psr/container": "^1.1 || ^2.0",
         "sulu/messenger": "^1.0",
-        "sulu/sulu": "dev-feature/content-resolver-path as 3.0.9",
+        "sulu/sulu": "^3.1@dev",
```
Prokyonn: Since the newest core release is 3.0.9, `^3.1@dev` can only resolve to `3.1.x-dev`, so this branch is installable only with dev minimum stability until 3.1 ships. Is that intended for the bundle's `3.0` branch, or should the branch track 3.1?

Prokyonn (commented):
The description is a single link, but the PR also moves `sulu/sulu` off the fork and adapts `ProductResolver` and the DQL casing assertion in `ProductRepositoryTest` to 3.1. Two sentences on that in the body would help anyone scanning for the titled change.

Prokyonn (approved):

PR #418, opened 2026-09-08
Add product attributes form

martinlagler:
Related to this PR: https://github.com/sulu/sulu/pull/9080

#### What's in this PR?

The bundle side of the `product_attributes` field type: the product's attribute values are no longer injected into the details form as server-side fields, the admin loads a `product_attributes` form per product family instead.

- `config/forms/Product/product_attributes.xml`: new, empty form the visitor fills.
- `ProductAttributesFormMetadataVisitor` is keyed on `product_attributes` and reads the family from the metadata options (`productFamily`, or `product` resolved to its family, plus `variant` for the axis attributes). One section per attribute group, one field per attribute named `attribute_<id>`. Always non-cacheable, since the response depends on family configuration that changes without the URL changing.
- `ProductAttributesSchemaFormMetadataVisitor` is new and keyed on the product forms themselves. It merges the attribute validation (required, min, max, step) into the JSON schema of `product_details`, one "if productFamily is X then attributes match X" branch per family, and of `product_variant`, the axis attributes of the parent's family, so the admin validates before saving.
- `AttributeFieldFactory` builds one field per attribute, the unit symbol is part of the label (`Length (mm)`), there is no unit field any more. `ProductVariantAttributeFormMetadataVisitor` is gone.
- `product_details.xml` and `product_variant.xml` declare `<property name="attributes" type="product_attributes"/>` inside an `attributes` section; the details section appears once a family is selected, the variant form passes `variant=true`.
- `ProductAdmin` no longer passes router attributes to the form metadata of the details view. The variants view keeps `['id' => 'parentId']`, which the variant schema reads.
- `ProductAttributesNormalizer` drops the `<id>_unit` keys.
- `ProductVariantController` always materialises `attributes`, so the required check runs for an untouched new variant. Create and update share `buildData()`, so this applies to `PUT` as well. `postAction` answers 422 / 400 like `putAction`, and `ProductController::postAction` now does the same instead of letting the exception surface as a 500.
- `ProductDetailsDataMapper`: dropped an `instanceof` that the locale guard above it has already decided.
- `ext-filter` moves from `require-dev` to `require`, since `ProductAttributesFormMetadataVisitor` uses `filter_var()`.
- Translations: `sulu_product.last_modified_or_authored` closes a de/en gap. The attribute container's own labels live in the core catalogue as `sulu_admin.*` keys.
- Tests: visitor and factory tests rewritten, `ProductAttributesMetadataTest` hits `/admin/metadata/form/product_attributes` over HTTP (the test app now imports the AdminBundle routing under `/admin`), product and variant controller tests for the 422 and 400 answers.

Needs a sulu release with sulu/sulu#9080 to render anything.


src/Infrastructure/Sulu/Admin/ProductAttributesFormMetadataVisitor.php
```diff
@@ -16,98 +16,96 @@
...
-        if (!\is_string($id)) {
-            return;
-        }
-
-        $family = $this->productFamilyRepository->findOneBy(
-            ['productUuid' => $id],
-            [ProductFamilyRepositoryInterface::GROUP_SELECT_PRODUCT_FAMILY_FORM => true],
-        );
+        // The response depends on family configuration, which changes without the URL changing.
+        $formMetadata->setCacheable(false);
```
Prokyonn, line 56: why is this not cacheable? I think the main issue is, that the admin ui has to be hard reloaded with the cache right? We already have this in some other cases, we should probably discuss this before the final release, depends also a bit on the performance impacts this has

src/Infrastructure/Sulu/Content/Normalizer/ProductAttributesNormalizer.php
```diff
@@ -52,15 +52,7 @@ public function enhance(object $object, array $normalizedData): array
...
-                $attribute = $familyAttribute->getAttribute();
-                $attributesMap[$attribute->getId()] = null;
-
-                $config = $attribute->getConfig();
-                $measurementFamily = $config['measurementFamily'] ?? null;
-                $unit = $config['unit'] ?? null;
-                if (\is_string($measurementFamily) && \is_string($unit)) {
-                    $attributesMap[$attribute->getId() . '_unit'] = $unit;
-                }
+                $attributesMap[$familyAttribute->getAttribute()->getId()] = null;
```
Prokyonn, line 55: null for each entry? is this intentional?

src/UserInterface/Controller/Admin/ProductController.php
```diff
@@ -162,8 +162,15 @@ public function postAction(Request $request): Response
         );
 
         $message = new CreateProductMessage($data);
-        /** @var ProductInterface $product */
-        $product = $this->handle(new Envelope($message, [new EnableFlushStamp()]));
+
+        try {
```
Prokyonn, line 174: **1.** The variant controller now always passes `attributes` to the mapper, because `ProductAttributesDataMapper` returns early when the key is missing. The product create path doesn't, so the required check is skipped here. With a family that has a required shared attribute:

- `POST /admin/api/products.json` without `attributes` → `201`
- the same request with `"attributes": {}` → `422`

`testPostWithMissingRequiredAttributeReturns422` sends `[$attributeId => null]`, so the missing key isn't covered. Could `postAction` default `attributes` to `[]` like `ProductVariantController::buildData()`, with a test that omits the key? Whether `putAction` should do the same is your call, the variant controller now applies it to PUT too.

src/Infrastructure/Sulu/Admin/ProductAttributesSchemaFormMetadataVisitor.php
```diff
@@ -0,0 +1,139 @@
...
+            $formMetadata->setSchema($formMetadata->getSchema()->merge($schema));
+        }
+    }
+
+    private function buildDetailsSchema(string $locale): ?SchemaMetadata
+    {
+        $select = [ProductFamilyRepositoryInterface::GROUP_SELECT_PRODUCT_FAMILY_FORM => true];
+        $branches = [];
+
+        foreach ($this->productFamilyRepository->findBy([], $select) as $family) {
```
Prokyonn: **2.** This adds a branch for every family to every `product_details` response, and the response is `no-store`, so it is rebuilt on each product open. Locally, 51 families with 20 attributes each take the metadata from 19 KB to 384 KB (0.07 s to 0.22 s), and the admin compiles that schema each time.

On the edit view the family can't change anymore (`disabledCondition="!!id"`), so keeping `addRouterAttributesToFormMetadata(['id'])` on the details view would let this visitor emit only the product's own family. Only the add form needs all branches.

src/Infrastructure/Sulu/Admin/AttributeFieldFactory.php
```diff
@@ -141,6 +151,7 @@ private function cloneFieldWithName(FieldMetadata $template, string $name): Fiel
         $field->setMaxOccurs($template->getMaxOccurs());
         $field->setSpaceAfter($template->getSpaceAfter());
         $field->setOnInvalid($template->getOnInvalid());
+        $field->setRequired($template->isRequired());
```
Prokyonn: **3.** `buildField()`, the only caller, overwrites this right after cloning with `$field->setRequired($familyAttribute->isRequired())`.

```suggestion
```

src/Infrastructure/Sulu/Admin/AttributeFieldFactory.php
```diff
@@ -16,42 +16,68 @@
...
- * {@see ProductAttributeFormMetadataVisitor} (the product's Details tab) and
- * {@see ProductVariantAttributeFormMetadataVisitor} (the variant overlay), so the field-cloning/template
- * resolution logic is not duplicated across visitors.
+ * Builds the admin field metadata for a single {@see ProductFamilyAttributeInterface}.
  *
  * @internal
  */
 class AttributeFieldFactory
 {
+    // The admin's ProductAttributesRenderer strips this prefix to recover the attribute id.
```
Prokyonn: **4.** The prefix is stripped in `ProductAttributes.js` (`handleChange`, `handleFinish`), the renderer only ever sees the prefixed names.

```suggestion
    // The admin's ProductAttributes container strips this prefix to recover the attribute id.
```

translations/admin.en.json
```diff
@@ -45,6 +45,7 @@
     "sulu_product.published": "Published",
     "sulu_product.not_published": "Not Published",
     "sulu_product.route": "Route",
+    "sulu_product.last_modified_or_authored": "Last modified or authored",
```
Prokyonn: **5.** I can't find `sulu_product.last_modified_or_authored` used anywhere, neither in the bundle nor in sulu/sulu. Remove it from `admin.de.json` instead?

Prokyonn (approved):
Looks good, most of the comments can be done in a follow up and maybe need internal discussion / metrics regarding the caching

src/Infrastructure/Sulu/Content/Normalizer/ProductAttributesNormalizer.php
martinlagler: Yes, intentional, and it is about where the validation error lands. required in JSON schema only checks that the key exists, so with a null present the value fails the per-field type check and the error is reported on attributes/<id>, which the container maps onto that row. 

Without the key, ajv reports required on the attributes object instead, and ProductAttributes.js returns early from its errors getter on a keyword error, so the save is blocked with no field marked.

src/Infrastructure/Sulu/Admin/ProductAttributesFormMetadataVisitor.php
martinlagler: Will be done in a follow up

PR #419, opened 2026-09-08
Add own URLs for Product Variants and Simple Products avoid URL for Product with Variant ( main product )

### LINKED ISSUE #415: Url / Route Handling for Variants
The target is that we want to be more flexible about URL handling for variants.

Sometimes variants may have there own URL sometimes Variants should not.

Currently discussion with Thomas goes more about a Dev Config

Maybe something like:

```yaml
sulu_product:
    variant_url_strategy: query | url
```

 - `query`: /products/nxasdf?variant=xyz
 - `url`: /products/nxasdf-xyz

Current Ruleset:

 - Simple Product always the main product has the URL
 - Product with Variant it depends on the above config

Open questions:

 - We maybe don't want that the variant can a full edit URL and so maybe the parent URL (prefix) still live on the main product.

```yaml
sulu_product:
    variant_url_strategy: query | url | child_url
```

 - `query`: /products/nxasdf?variant=xyz
 - `url`: /products/nxasdf-xyz
 - `child_url`: /products/nxasdf/xyz

In our current project URL schema more goes about Main Product = No URL and variant has URL inside its main category / family.

# Research Results

As a first I wanted todo a research how other Systems handling URLs, concentrate on the basic installations.

 - [Shopify](#shopify)
   - all products under `/products/...` slug edit possible
   - a URL history from old to new
   - variants URL just query strings, subscription change no URL effects, no edit
   - breadcrumb only via additional plugin and there fixed not dynamic
 - [WooCommerce](#WooCommerce)
   - all products under `/product/...` slug edit possible
   - a URL history from old to new
   - no variants URL always no variants selected
   - breadcrumb is always the first categories
 - [Sylius](#Sylius)
   - all products under `/products/...` slug edit possible
   - no URL history
   - no variants URL always first variant selected
   - breadcrumb is always the main taxon
 - [Shopware](#Shopware)
   - products has there own URL full editable URL
   - variants has there own URL full editable URL
   - redirect from old variant to new variant works
   - default variants URL is something like `/<slug>/<code>.<variant_increased_nr>`
   - main product URL `/<slug>/<code> change` it has no effect on variant URL
   - redirect from old variant to new variant works
   - redirect from old main product to new main product does not work
   - breadcrumb is always the seo category, if not defined first which feels like random 
   - links from category pages are to the first variant

## Shopify

Shopify have products and it have collections, in case of shopify the collection does not effect the product Url.

It seems that all Products live under a simple `/products/` prefix and it is also the one shown on the collection. Currently no Breadcrumb or something else.

Image

### Shopify Breadcrumb Handling

No breadcrumb inside the default Shopify Shop Theme. Still interesting `/<collection_url>/<product_url>` will redirect you to `/<product_url>` e.g.: 

 - https://sulu-dev-shop-plus.myshopify.com/collections/hydrogen/products/gift-card-1234
    - https://sulu-dev-shop-plus.myshopify.com/products/gift-card-1234

This redirect only work for active URLs not for history URLs.

Plugins for Shopify like Breadcrumbs & Categories still use same routing: https://breadcrumbs-and-categories.myshopify.com/products/nike-mid-century-oak-kids-beds-stained-olive

### Shopify Variants Urls

In case of variants there exists kind of two ways one which effects the URL via a query string e.g.: 

Image

The "Subscription" Part has no effects on URL:

Image

The variant part can not be edited is is mostly a internal Snowflake ID of shopify.

### Shopify Redirect Handling

There exist no tree URL in Shopify so there is so no update required to update child routes, it directly shows you if you want todo a redirect from old to new URL and let manager decide what they want todo.

Image

## WooCommerce

WooCommerce have products and it have categories, in case of WooCommerce the categories does not effect the product Url.

It seems that all Products live under a simple `/product/` prefix. Breadcrumb is the first category (kind of random feeling)

Image

### WooCommerce Breadcrumb Handling

A breadcrumb to the first category exists. But there is no special URL for that:

### WooCommerce Variants Urls

Variants / variant changes has no directly effects to the URL in WooCommerce default theme.

### WooCommerce Redirect Handling

There exist no tree URL WooCommerce Shopify so there is so no update required to update child routes, only redirect from old to new URL.

## Sylius

Sylius have products and it have taxons, in case of Sylius the taxons does not effect the product Url.

It seems that all Products live under a simple `/products/` prefix. A product has main taxon and it is also the one shown on the collection as a breadrumb.

Image

### Sylius Breadcrumb Handling

Only the main taxon effects the breadcrumb. No taxon specific product URLs.

### Sylius Variants Urls

No variant URLs by default, preselect first variant.

### Sylius Redirect Handling

No redirect handling, old Slug not longer reachable.

## Shopware

Shopware have products and it have categories, in case of Sylius the categories does not effect the product Url.

It seems that all have there own URL and even variants have there own FullEditable URL. A product has seo category and it is also the one to show the breadcrumb else it uses the first category.

Image

### Shopware Breadcrumb Handling

Choosen SEO Category or First category if not defined:

Image

### Shopware Variants Urls

Variants have there complete own URL the category can not be changed and is greyed out so used from the main product:

Image

### Shopware Redirect Handling

Redirects are handled, but it feels kid of strange as editing the Main Product URL not changes Variants URLs.

alexander-schranz:
fixes #415 ( just simple case )

The route field of a product no longer lives in the product template but in the `product_details`
form and the `product_variant` overlay. Its field type and its params come from a new
`sulu_product.route` config, applied by a form metadata visitor, so a project configures the route
once instead of per template:

```yaml
sulu_product:
    route:
        type: route # default, e.g. "page_tree_route" for a route below a page
        params:
            route_schema: "/products/{implode('-', object)}" # default
```

`params` takes any key the configured field type understands and forwards it to the field as-is.
A configured param wins over the one the form declares and over the default of the same key, so a
project states only what it changes.

The same visitor adds the field invisibly to every product template, because
[`RoutableDataMapper`](https://github.com/sulu/sulu/blob/3.0/packages/content/src/Application/ContentDataMapper/DataMapper/RoutableDataMapper.php#L75)
reads the route property off the template metadata. That field carries the skip tag of
`TemplateDataMapper`, so the slug stays with the route entity.

A product with variants is reached through its variants and owns no route of its own:

* the field is hidden on `product_details` for that type
* `ProductWithVariantsRouteGuard`, a doctrine `onFlush` listener, drops a route that reaches such a
  dimension content programmatically and detaches one the same flush would insert
* on the variant overlay the route is mandatory instead, set through the visitor together with a
  merged schema constraint, because the schema is built before the visitors run

What to check: the config reaches both forms, the URL generates from the title, a saved product
gets its route, and a product with variants gets none.

## Out of scope

A variant still resolves to its parent URL plus `?variant=`. Replacing
that with the real variant URLs and removing `sulu_product.variant_query_parameter` belongs in a
separate pull request, this one keeps the current behaviour.

Let this follow up on the resolving part handled by @martinlagler or @Prokyonn .

---

I also update the Form so Mandatory fields are at the top, the type has the most effects and need to be the first field:

Bildschirmfoto 2026-09-09 um 18 31 51

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,17 @@ sulu_product:
     route:
         type: route # default, e.g. "page_tree_route" for a route below a page
         params:
-            route_schema: "/products/{implode('-', object)}"
+            route_schema: "/products/{implode('-', object)}" # default
 ```
 
 `params` takes any key the configured field type understands and forwards it to the field as-is;
 `route_schema` is the one the `route` field reads to generate the URL out of the fields tagged
 `sulu.rlp.part`. A param configured here wins over the same param declared in the form XML. Both
 forms get the same type and the same params.
+
+`route_schema` defaults to the value above, so a product URL starts with `/products/` without any
+configuration. A project overrides that key or adds params of its own, and the default stays for
+every key the project does not set.
 
 The same field is added invisibly to every product template, because `RoutableDataMapper` of the
 content package reads the route property off the template metadata. A template declaring its own
--- a/README.md
+++ b/README.md
@@ -29,21 +29,10 @@ content package reads the route property off the template metadata. A template d
 
 ## Variant URLs
 
-A variant owns no route of its own. Referenced from a page, it resolves to its parent's URL plus a
-query parameter carrying the variant code, so `/products/cable` with code `XY-2` resolves to
-`/products/cable?variant=XY-2`.
-
-The bundle only writes that URL — nothing reads the parameter back off the request. A project
-selects the variant itself, which makes the key a contract between both sides: change it under
-`sulu_product.variant_query_parameter` and the reading side has to match by hand.
-
-```yaml
-sulu_product:
-    variant_query_parameter: 'variant' # default
-```
-
-A variant whose parent has no published route in the requested locale resolves without a `url`,
-because an empty string would point at the site root.
+A variant owns its route: the URL field is mandatory on the variant overlay, so every variant
+carries an address of its own. A product with variants owns none, it is reached through its
+variants, which is why the field is hidden for that type and the route guard drops one that
+reaches such a product programmatically.
 
 ## Association form overrides
 
--- a/README.md
+++ b/README.md
@@ -19,6 +19,10 @@ sulu_product:
 `sulu.rlp.part`. A param configured here wins over the same param declared in the form XML. Both
 forms get the same type and the same params.
 
+The same field is added invisibly to every product template, because `RoutableDataMapper` of the
+content package reads the route property off the template metadata. A template declaring its own
+`url` property keeps it and only receives the configured type and params.
+
 ## Variant URLs
 
 A variant owns no route of its own. Referenced from a page, it resolves to its parent's URL plus a
--- a/README.md
+++ b/README.md
@@ -1,5 +1,24 @@
 # Sulu Product Bundle
 
+## Product route
+
+The route field of a product lives in the `product_details` form and in the `product_variant`
+overlay, not in the product template. Its field type and its params are configured once for the
+whole project, so a form does not repeat them:
+
+```yaml
+sulu_product:
+    route:
+        type: route # default, e.g. "page_tree_route" for a route below a page
+        params:
+            route_schema: "/products/{implode('-', object)}"
+```
+
+`params` takes any key the configured field type understands and forwards it to the field as-is;
+`route_schema` is the one the `route` field reads to generate the URL out of the fields tagged
+`sulu.rlp.part`. A param configured here wins over the same param declared in the form XML. Both
+forms get the same type and the same params.
+
 ## Variant URLs
 
 A variant owns no route of its own. Referenced from a page, it resolves to its parent's URL plus a
--- a/config/forms/Product/product_details.xml
+++ b/config/forms/Product/product_details.xml
@@ -14,6 +14,16 @@
             <params>
                 <param name="headline" value="true"/>
             </params>
+
+            <tag name="sulu.rlp.part"/>
+        </property>
+
+        <property name="url" type="route" colspan="12">
+            <meta>
+                <title>sulu_product.url</title>
+            </meta>
+
+            <tag name="sulu.rlp"/>
         </property>
 
         <property name="code" type="text_line" colspan="6">
--- a/config/forms/Product/product_details.xml
+++ b/config/forms/Product/product_details.xml
@@ -18,7 +18,7 @@
             <tag name="sulu.rlp.part"/>
         </property>
 
-        <property name="url" type="route" colspan="12">
+        <property name="url" type="route">
             <meta>
                 <title>sulu_product.url</title>
             </meta>
--- a/config/forms/Product/product_details.xml
+++ b/config/forms/Product/product_details.xml
@@ -6,7 +6,7 @@
     <key>product_details</key>
 
     <properties>
-        <property name="type" type="single_select" mandatory="true" colspan="4" disabledCondition="!!id or url" >
+        <property name="type" type="single_select" mandatory="true" colspan="4" disabledCondition="!!id or url">
             <!-- as soon as an URL generated we can not allow to switch back as emptying an URL is not possible -->
             <meta>
                 <title>sulu_product.product_type</title>
--- a/config/forms/Product/product_details.xml
+++ b/config/forms/Product/product_details.xml
@@ -6,45 +6,8 @@
     <key>product_details</key>
 
     <properties>
-        <property name="title" type="text_line" mandatory="true" colspan="12">
-            <meta>
-                <title>sulu_admin.name</title>
-            </meta>
-
-            <params>
-                <param name="headline" value="true"/>
-            </params>
-
-            <tag name="sulu.rlp.part"/>
-        </property>
-
-        <property name="url" type="route">
-            <meta>
-                <title>sulu_product.url</title>
-            </meta>
-
-            <tag name="sulu.rlp"/>
-        </property>
-
-        <property name="code" type="text_line" colspan="6">
-            <meta>
-                <title>sulu_product.code</title>
-            </meta>
-        </property>
-
-        <property name="status" type="single_select" colspan="6">
-            <meta>
-                <title>sulu_product.status</title>
-            </meta>
-        </property>
-
-        <property name="productFamily" type="single_product_family_selection" mandatory="true" colspan="6" disabledCondition="!!id" multilingual="false">
-            <meta>
-                <title>sulu_product.product_family</title>
-            </meta>
-        </property>
-
-        <property name="type" type="single_select" mandatory="true" colspan="6" disabledCondition="!!id">
+        <property name="type" type="single_select" mandatory="true" colspan="4" disabledCondition="!!id or url" >
+            <!-- as soon as an URL generated we can not allow to switch back as emptying an URL is not possible -->
             <meta>
                 <title>sulu_product.product_type</title>
             </meta>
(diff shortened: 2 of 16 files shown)


translations/admin.de.json
```diff
@@ -32,6 +32,7 @@
     "sulu_product.product_type_product_with_variants": "Produkt mit Varianten",
     "sulu_product.product_type_variant": "Variante",
     "sulu_product.code": "Code",
+    "sulu_product.url": "Adresse",
```
alexander-schranz, line 36: maybe a translation which we better should move to the core admin bundle.

src/Infrastructure/Doctrine/EventListener/ProductWithVariantsRouteGuard.php
```diff
@@ -0,0 +1,68 @@
...
+            $metadata = $entityManager->getClassMetadata($entity::class);
+            if ($unitOfWork->isScheduledForInsert($entity)) {
+                $unitOfWork->computeChangeSet($metadata, $entity);
+            } else {
+                $unitOfWork->recomputeSingleEntityChangeSet($metadata, $entity);
+            }
+
+            // a route the same flush would insert has no owner left
+            if ($unitOfWork->isScheduledForInsert($route)) {
+                $entityManager->detach($route);
```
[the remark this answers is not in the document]
alexander-schranz: A route is created together with the product dimension content and never handed from one content to another, so a product with variants cannot end up owning a persisted route. Nulling the association and detaching a route the same flush would insert covers the case.

src/Infrastructure/Sulu/Admin/ProductRouteFormMetadataVisitor.php
```diff
@@ -0,0 +1,122 @@
...
+        $this->applyRouteConfig($routeField);
+
+        if (ProductInterface::FORM_KEY_VARIANT !== $formMetadata->getKey()) {
+            return;
+        }
+
+        $routeField->setRequired(true);
+        $formMetadata->setSchema($formMetadata->getSchema()->merge(new SchemaMetadata([
+            new PropertyMetadata(self::FIELD_NAME, true, new StringMetadata(1)),
+        ])));
```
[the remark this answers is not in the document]
alexander-schranz: Removed the schema merge. The visitor only sets `required` on the field now, how a required field is enforced belongs to the core metadata, not to this visitor.

src/Infrastructure/Doctrine/EventListener/ProductWithVariantsRouteGuard.php
```diff
@@ -0,0 +1,68 @@
...
+{
+    public function onFlush(OnFlushEventArgs $eventArgs): void
+    {
+        $entityManager = $eventArgs->getObjectManager();
+        $unitOfWork = $entityManager->getUnitOfWork();
+
+        $entities = [
+            ...$unitOfWork->getScheduledEntityInsertions(),
+            ...$unitOfWork->getScheduledEntityUpdates(),
+        ];
```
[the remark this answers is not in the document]
alexander-schranz: The type cannot be changed after creation, the form disables the field once an id or a url exists. An import that changes the type has to handle the URLs itself.

src/Infrastructure/Symfony/HttpKernel/SuluProductBundle.php
```diff
@@ -371,6 +393,11 @@ public function loadExtension(array $config, ContainerConfigurator $container, C
         $variantQueryParameter = $config['variant_query_parameter'] ?? 'variant';
         $builder->setParameter('sulu_product.variant_query_parameter', $variantQueryParameter);
 
+        /** @var array{type: string, params: array<string, scalar|null>} $route */
+        $route = $config['route'] ?? ['type' => 'route', 'params' => []];
+        $builder->setParameter('sulu_product.route.type', $route['type']);
+        $builder->setParameter('sulu_product.route.params', $route['params']);
```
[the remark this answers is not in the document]
alexander-schranz: Fixed. The params node merges the defaults into the configured ones through `validate()->always()`, so a project that sets a single param keeps the default `route_schema`.

README.md
```diff
@@ -1,5 +1,32 @@
 # Sulu Product Bundle
 
+## Product route
+
+The route field of a product lives in the `product_details` form and in the `product_variant`
+overlay, not in the product template. Its field type and its params are configured once for the
```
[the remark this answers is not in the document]
alexander-schranz: Updated. The Variant URLs section now describes the route a variant owns and no longer mentions `?variant=`, the resolver change is a follow up.

README.md
```diff
@@ -1,22 +1,38 @@
...
+configuration. A project overrides that key or adds params of its own, and the default stays for
+every key the project does not set.
+
+The same field is added invisibly to every product template, because `RoutableDataMapper` of the
+content package reads the route property off the template metadata. A template declaring its own
+`url` property keeps it and only receives the configured type and params.
+
+## Variant URLs
+
+A variant owns its route: the URL field is mandatory on the variant overlay, so every variant
```
Prokyonn, line 32: This section drops the `sulu_product.variant_query_parameter` documentation, but the option is untouched and still builds `?variant=` in `ProductController.php:139` and `ProductResourceLoader.php:117`, and the description says variant URL resolution is out of scope.

src/Infrastructure/Symfony/HttpKernel/SuluProductBundle.php
```diff
@@ -370,6 +401,11 @@ public function loadExtension(array $config, ContainerConfigurator $container, C
         $variantQueryParameter = $config['variant_query_parameter'] ?? 'variant';
         $builder->setParameter('sulu_product.variant_query_parameter', $variantQueryParameter);
 
+        /** @var array{type: string, params: array<string, scalar|null>} $route */
+        $route = $config['route'] ?? ['type' => 'route', 'params' => []];
```
Prokyonn: `addDefaultsIfNotSet()` guarantees this key, so the `??` cannot trigger, and its fallback would drop `route_schema` if it ever did. Reading `$config['route']` directly says what actually happens.

tests/Functional/Integration/ProductControllerTest.php
```diff
@@ -200,6 +202,30 @@ public function testPost(): void
         $this->assertNotEmpty($id);
     }
 
+    public function testPostProductWithVariantsWithoutUrlCreatesNoRoute(): void
```
Prokyonn, line 205: This posts no `url`, and core's `RoutableDataMapper` returns early when the payload carries no route data, so no route is ever created and the test passes with the guard deleted. The same test posting a product with variants *with* a url would exercise the guard, and it would have caught the changeset crash the last commit fixed.

src/Infrastructure/Sulu/Admin/ProductRouteFormMetadataVisitor.php
```diff
@@ -0,0 +1,116 @@
...
+            return;
+        }
+
+        foreach ($formMetadata->getForms() as $form) {
+            $routeField = $form->getItems()[self::FIELD_NAME] ?? null;
+
+            if (!$routeField instanceof FieldMetadata) {
+                $routeField = new FieldMetadata(self::FIELD_NAME);
+                $routeField->setVisibleCondition('false');
+                $this->addTag($routeField, self::RESOURCE_LOCATOR_TAG);
```
Prokyonn, line 84: The injected field gets `sulu.rlp` and the skip tag, but the template property it replaces also carried `<tag name="sulu.search.field" role="url"/>`, which `title` still has right beside it. A project that removes its own `url` property the way the test template does loses that role, so either the visitor should add it too or the README should name the loss.

config/forms/Product/product_details.xml
```diff
@@ -6,56 +6,71 @@
     <key>product_details</key>
 
     <properties>
-        <property name="title" type="text_line" mandatory="true" colspan="12">
+        <property name="type" type="single_select" mandatory="true" colspan="4" disabledCondition="!!id or url" >
```
Prokyonn: Trailing space before the bracket: `disabledCondition="!!id or url" >`.

README.md
alexander-schranz: yes that I let as follow up for the resolving the remove that part.

src/Infrastructure/Sulu/Admin/ProductRouteFormMetadataVisitor.php
alexander-schranz: make sense to add it

src/Infrastructure/Symfony/HttpKernel/SuluProductBundle.php
alexander-schranz: make sense to remove it then.

src/Infrastructure/Doctrine/EventListener/ProductWithVariantsRouteGuard.php
```diff
@@ -0,0 +1,64 @@
...
+            $route = $entity->getRoute();
+            if (!$route instanceof Route) {
+                continue;
+            }
+
+            if (!$entity->getResource()->isType(ProductInterface::TYPE_PRODUCT_WITH_VARIANTS)) {
+                continue;
+            }
+
+            $entity->removeRoute();
```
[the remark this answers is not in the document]
Prokyonn: resolving is out of scope for this PR, needs a followup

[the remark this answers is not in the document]
alexander-schranz: resolving is out of scope and need be tackled in upcoming PR

src/Infrastructure/Sulu/Admin/ProductRouteFormMetadataVisitor.php
```diff
@@ -0,0 +1,123 @@
...
+            return;
+        }
+
+        $this->applyRouteConfig($routeField);
+
+        if (ProductInterface::FORM_KEY_VARIANT !== $formMetadata->getKey()) {
+            return;
+        }
+
+        $routeField->setRequired(true);
```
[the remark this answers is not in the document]
alexander-schranz: need to be false its a static method and we are dynamic now here.

Prokyonn (approved):

PR #420, opened 2026-09-09
Add search

martinlagler:
Indexes products into Sulu's shared `website` index, so they are found next to pages and articles, and adds filter fields for a project's own catalogue search.

## What changed

- **Documents:** one document per published leaf, a product without variants or a variant. A product with variants gets none. A variant document uses its own route and its parent's webspaces. The product code is searchable through `content`. The admin index keeps the parent and skips variants.
- **Filter fields** under the object field `product` (schema in `config/schemas/website.php`):
  - `productFamilyId`
  - `attributes_text_values`: one `<attributeKey>:<value>` entry per option key and text value
  - `attributes_numeric_values.<attributeKey>`: number and date attributes
- **Content:** family name, external identifier, short description and `Label: value` attribute texts, and the details image as fallback `mediaId`. A variant also gets its parent's values that are not variant-specific.
- **Schema:** `ProductSchemaLoader` appends a field per number or date attribute. A new such attribute (`AttributeCreatedEvent`) rebuilds the website index.
- **Queries:** attributes and values are loaded once per reindex run, so the query count does not grow with the products (`ProductReindexProviderQueryCountTest`).
- **Config:** the filter fields can be turned off:

  ```yaml
  sulu_product:
      search:
          website:
              additional_product_filters: false # default: true
  ```

- **`ProductResourceLoader`:** no longer builds a variant URL from the parent's route plus `?variant=`. Since #419 a variant has its own route, so a referenced variant links to it.

The bundle ships no catalogue controller. The README shows how to build one on SEAL's `EngineInterface`:

```php
$engine->createSearchBuilder('website')
    ->addFilter(Condition::equal('resourceKey', ProductInterface::RESOURCE_KEY))
    ->addFilter(Condition::equal('product.attributes_text_values', 'colour:black'))
    ->addFilter(Condition::greaterThanEqual('product.attributes_numeric_values.weight', 20.0));
```

## Please check

- A new number or date attribute reaches the live index only after `bin/console cmsig:seal:reindex --index website --drop`. Until then, a strict engine such as Elasticsearch rejects products that carry a value for it. This currently happens automatically when the attribute is created: `WebsiteProductSchemaListener` dispatches `RebuildWebsiteIndexMessage`, which drops and refills the whole `website` index, pages and articles included.
- Nested filters need an adapter that resolves dotted paths (Elasticsearch, Loupe). The memory adapter does not.
- `ProductReindexProviderQueryCountTest` is skipped on the lowest dependencies, because DoctrineBundle 2.6 has no `doctrine.debug_data_holder`.
- The website `ProductController` still selects a variant via `?variant=` (`selectedVariant`). Since #419 that is probably dead as well, but projects read `selectedVariant`, so it is left for a separate change.


src/Infrastructure/Sulu/Search/Visitor/BatchAwareReindexEnhancerInterface.php
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * An enhancer that needs data for the whole batch loads it once here instead of per document.
+ */
+interface BatchAwareReindexEnhancerInterface
+{
+    /**
+     * @param array<int, array<string, mixed>> $rows the query rows of the batch about to be enhanced
+     */
+    public function prepareBatch(array $rows): void;
+}
```
alexander-schranz: Looks unexpected extension, we do not have this for Pages, Articles, ... so if there is an issue we need resolve we need a general solution or handling things inside the enhancer instead of a new public interface.

src/Domain/Event/ProductRemovedEvent.php
```diff
@@ -69,4 +70,15 @@ public function getAllLocales(): ?array
...
+    /**
+     * Uuids of the variants the removal takes with it, as they fire no event of their own. The
+     * removed product itself is not part of the list.
+     *
+     * @return string[]
+     */
+    public function getRelatedProductIds(): array
+    {
+        return $this->context['relatedIds'] ?? [];
+    }
```
alexander-schranz, line 83: should we not call it `getVariantUuids` here?

src/Domain/Event/ProductRemovedEvent.php
```diff
@@ -21,7 +21,8 @@ class ProductRemovedEvent extends DomainEvent
 {
     /**
      * @param array{
-     *     locales?: string[]
+     *     locales?: string[],
+     *     relatedIds?: string[]
```
alexander-schranz: related sounds like associations, think this should more be `variantUuids`

src/Infrastructure/Sulu/Search/ProductIndex.php
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+namespace Sulu\Product\Infrastructure\Sulu\Search;
+
+use Sulu\Product\Domain\Model\ProductInterface;
+
+/**
+ * Names of the shared website index, of the product object field inside it and of its documents.
+ */
+final class ProductIndex
```
alexander-schranz: did we have such class for pages and articles? I more woudl put thi as private or when requires as internal static methods on the `WebsiteProductReindexProvider`.

src/Infrastructure/Sulu/Search/WebsiteProductReindexProvider.php
alexander-schranz, line 1: This Class should be similar to the `WebsiteArticleReindexProvider` and all `Product` related additional fields (attributes, ...) should be handled in an own service implementing `WebsiteProductReindexProviderEnhancerInterface`, e.g.: `WebsiteProductDetailsReindexProviderEnhancer`.

That allows to disables the whole attributes indexing, ... for projects which may not need such filters just a basic search. And we make sure that our Enhancer pattern fulfilles all requirements to even do this kind of things.

config/schemas/website.php
```diff
@@ -0,0 +1,25 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+use CmsIg\Seal\Schema\Field;
+use CmsIg\Seal\Schema\Index;
+
+return new Index('website', [
+    'product' => new Field\ObjectField('product', [
+        'code' => new Field\TextField('code', searchable: false, filterable: true),
```
alexander-schranz: The `code` does not make sense as a filter, as you will never filter on it as it is `unique`.

As its only kind of a searchable part the Reindex Provider should add it to the `content` array.

If you need it for displaying it we could store it into the unstructured [metadata](https://github.com/sulu/sulu/blob/6a7aee4630b542034ef22c5ff6650a62e1fcfdfe/packages/search/config/schemas/website.php#L26) field: would wrap it into a object e.g. ( this should live inside a project as most project only require the variant title not the variant code / product title not product code ) :

```php
[
    'metadata' => [
         'product' => ['code' => ...],
    ],
],
```

Inside `metadata` fields are stored which don't need be filterable, but maybe needed to display data. If you not requiring it now I would even remove it.

src/Infrastructure/Sulu/Search/WebsiteProductReindexProvider.php
```diff
@@ -21,23 +21,54 @@
 use Sulu\Product\Domain\Model\ProductDimensionContentAdditionalWebspace;
 use Sulu\Product\Domain\Model\ProductDimensionContentInterface;
 use Sulu\Product\Domain\Model\ProductInterface;
+use Sulu\Product\Infrastructure\Sulu\Search\Visitor\BatchAwareReindexEnhancerInterface;
 use Sulu\Product\Infrastructure\Sulu\Search\Visitor\WebsiteProductReindexProviderEnhancerInterface;
 
 /**
+ * One document per published leaf: a product without variants, or a variant. A product that has
+ * variants is represented by them, so it gets no document of its own. A variant is shown on its
+ * parent's page, so its document carries the parent's slug and webspaces.
```
alexander-schranz: URL / Slug is completly unrelated and should not be mention inside this comment.

src/Infrastructure/Sulu/Search/WebsiteProductReindexProvider.php
```diff
@@ -123,48 +135,115 @@ public function provide(ReindexConfig $reindexConfig): \Generator
...
+            ProductIndex::FIELD => [
+                'code' => (string) $row['code'],
+                'type' => $row['type'],
+                'status' => $row['status'],
+                'productFamilyId' => (string) $row['productFamilyId'],
+                'productFamilyName' => (string) $row['productFamilyName'],
+                'changedAt' => $row['changed']->format('c'),
+                ProductIndex::TEXT_VALUES_FIELD => [],
+                ProductIndex::NUMERIC_VALUES_FIELD => [],
+                'attributes' => [],
```
alexander-schranz: Fields are costly, do not create unnecessary fields.

Also you can not create fields which are not part of the schema

```suggestion
                'productFamilyId' => (string) $row['productFamilyId'],
                ProductIndex::TEXT_VALUES_FIELD => [],
                ProductIndex::NUMERIC_VALUES_FIELD => [],
```

src/Infrastructure/Sulu/Search/WebsiteProductReindexProvider.php
```diff
@@ -123,48 +135,115 @@ public function provide(ReindexConfig $reindexConfig): \Generator
...
+                . ' AND unlocalizedDimensionContent.locale IS NULL'
+                . ' AND unlocalizedDimensionContent.stage = dimensionContent.stage'
+                . ' AND unlocalizedDimensionContent.version = dimensionContent.version',
+            )
             ->leftJoin('dimensionContent.route', 'route')
-            ->select('IDENTITY(dimensionContent.product) AS productId')
-            ->addSelect('dimensionContent.authored')
-            ->addSelect('dimensionContent.changed')
-            ->addSelect('dimensionContent.title')
+            ->leftJoin('unlocalizedDimensionContent.productFamily', 'productFamily')
```
alexander-schranz: @chirimoya can we discuss this. I don't think that the relation from `Product` -> `ProductFamily` should be on Dimension. Its not changable after creating and so should live on the `Product` entity.

alexander-schranz (changes requested):
We should introduce a config like:

```yaml
sulu_product:
    search:
        website:
            additional_product_filters: true
```

If disabled:

 - the `config/schema/website.php` is registered in the prepend extension.
 - the website reindex provider enhancer which handles attributes + family isn't registered as a service
 - the schema loader for this fields is not registered

README.md
```diff
@@ -18,6 +18,95 @@ sulu_product:
...
+use Sulu\Product\Domain\Model\ProductInterface;
+
+$result = $engine->createSearchBuilder('website')
+    ->addFilter(Condition::equal('resourceKey', ProductInterface::RESOURCE_KEY))
+    ->addFilter(Condition::equal('locale', $locale))
+    ->addFilter(Condition::equal('webspaces', $webspaceKey))
+    ->addFilter(Condition::search($term))
+    ->addFilter(Condition::equal('product.attributes_text_values', 'colour:black'))
+    ->addFilter(Condition::greaterThanEqual('product.attributes_numeric_values.weight', 20.0))
+    ->addFacet(Facet::count('product.attributes_text_values'))
```
alexander-schranz: think `product.attributes_numeric_values.weight` would be the better example here:

```suggestion
    ->addFacet(Facet::count('product.attributes_numeric_values.weight'))
```

src/Infrastructure/Sulu/Search/Visitor/WebsiteProductDetailsReindexProviderEnhancer.php
```diff
@@ -0,0 +1,291 @@
...
+     *
+     * The product and its parent get a subquery each: with both in one `OR`, MySQL no longer uses
+     * the product index and scans every dimension content of the locale per row.
+     */
+    private function valuesSubquery(QueryBuilder $queryBuilder, string $alias, string $productExpression): string
+    {
+        $u = "'" . self::UNIT_SEPARATOR . "'";
+        $r = "'" . self::RECORD_SEPARATOR . "'";
+
+        $subquery = $queryBuilder->getEntityManager()->createQueryBuilder()
```
Prokyonn: What about instead of making this complicated subquery. Make one independent query that fetches all attributes (independent of the product) with the translations, save them inmemory in an array and in the enhanceDocument fill in the missing values from this array to the product.

src/Infrastructure/Sulu/Search/WebsiteProductSchemaListener.php
```diff
@@ -0,0 +1,46 @@
...
+    ) {
+    }
+
+    public function onAttributeCreated(AttributeCreatedEvent $event): void
+    {
+        if (!\in_array($event->getAttribute()->getType(), self::SCHEMA_TYPES, true)) {
+            return;
+        }
+
+        $this->messageBus->dispatch(new RebuildWebsiteIndexMessage());
```
Prokyonn, line 44: This rebuilds the whole shared `website` index (pages, articles and products) synchronously inside the admin request that creates the attribute, because `sulu_message_bus` handles the message in-process unless a project routes it elsewhere.

In the test environment (4 products, a few pages, Loupe) the `POST /admin/api/attributes` took about 2 s, and the public site search did not answer for 1.8 s while the index was being recreated. On Elasticsearch, `withDropIndex(true)` deletes the index first, so the whole site search is empty until the refill finishes. On a real site that refill takes longer than a request may run. If PHP's time or memory limit kills it halfway, the site search stays incomplete until someone runs the reindex command by hand. An editor adding a "Length" attribute should not be able to do that.

I'd remove the listener, the message and its handler, and keep what the README already says: after adding a number or date attribute, run `bin/console cmsig:seal:reindex --index website --drop` at a time you choose. If an automatic path is wanted later, it needs to run outside the request and avoid dropping the live index, which SEAL does not support yet (see the TODO in `RebuildWebsiteIndexMessageHandler`).

src/Infrastructure/Sulu/Search/Visitor/WebsiteProductDetailsReindexProviderEnhancer.php
```diff
@@ -0,0 +1,391 @@
...
+                ->select('IDENTITY(variant.parent)')
+                ->where('variant.uuid IN (:scope)')
+                ->getDQL();
+            $queryBuilder
+                ->andWhere('dimensionContent.product IN (:scope) OR dimensionContent.product IN (' . $parents . ')')
+                ->setParameter('scope', $this->scope);
+        }
+
+        /** @var iterable<array{productId: string, locale: string|null, attributeId: int, optionKey: string|null, number: float|null, text: string|null, variantSpecific: bool|null}> $rows */
+        $rows = $queryBuilder->getQuery()->toIterable();
```
Prokyonn, line 318: On a full reindex `scope` is null, so this loads every live attribute value of every product in every locale into one array, and keeps it for the whole run. Memory then grows with the catalogue: a quick measurement of just this array for 20,000 products with 15 attributes in 3 locale buckets comes to about 390 MB, before Doctrine's own overhead. That breaks `cmsig:seal:reindex` under a usual `memory_limit`, and it is what the attribute-triggered rebuild runs inside a web request.

It also has two side effects:
- The scope is read from the provider's `:id<n>` parameter names (line 117), a naming convention between two classes that nothing enforces.
- Two reindexes of the same products in one process (an import script publishing a product twice) reuse the first run's values, because the cache only resets when the scope changes.

Loading per batch avoids all three problems without a new interface. In `WebsiteProductReindexProvider::loadBatch()`, set `setFirstResult()`/`setMaxResults()` before the enhancers run. Then `enhanceQuery()` can take the batch's product and parent ids from a clone of the query builder and load the values for just those ids. That is one query per 100 products, memory stays bounded, and there is no parameter-name sniffing. Caching the attribute definitions for the whole run is fine, since they grow with the attributes, not the products.

README.md
```diff
@@ -80,6 +80,95 @@ switcher, format it with `sulu_product_format_attribute_value`:
...
+those of its parent that the family does not mark variant-specific. The same enhancer adds the product
+family name, external identifier, short description and attribute texts to the searchable `content`,
+and the details image where neither the template nor the excerpt has one.
+
+Option and text attributes need no field of their own, so adding one changes no schema. A number or
+date attribute does: `ProductSchemaLoader` reads the attribute table and appends its field to
+`attributes_numeric_values`. The live index only learns it when it is recreated, which is never
+automatic:
+
+    bin/console cmsig:seal:reindex --index website --drop
```
Prokyonn, line 124: This says the recreation "is never automatic", but `WebsiteProductSchemaListener` does it automatically on every new number or date attribute. If the listener goes (see my comment there), this paragraph is right as it stands. Otherwise it needs to describe the automatic rebuild.

Prokyonn (changes requested):
Thanks, the document model works well in the browser: the website search finds the variants with their own URLs, the parent has no document, the admin search keeps only the parent, and the README's filter and facet examples return the right products on Loupe.

Two things need to change before this can go in, both commented inline: creating a number or date attribute rebuilds the whole `website` index inside the admin request, and a full reindex loads the attribute values of the entire catalogue into memory at once.

Please also rebase onto 3.0. The red test jobs are not from this branch: its base (#430) was red on 3.0 itself, and #433 fixed that after you branched. The rebase conflicts in `tests/Functional/Content/ProductResolverAssociationsTest.php`. The test this PR adds there, `testAssociatedVariantLinksToItsOwnUrl`, covers `ProductResourceLoader`, which this PR no longer changes, so I'd drop it here. The description's `ProductResourceLoader` paragraph is outdated for the same reason. After the rebase, CI can actually vouch for the new search tests.

src/Infrastructure/Symfony/HttpKernel/SuluProductBundle.php
```diff
@@ -285,6 +287,20 @@ public function configure(DefinitionConfigurator $definition): void
...
                 ->end()
+                ->arrayNode('search')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->arrayNode('website')
+                            ->addDefaultsIfNotSet()
+                            ->children()
+                                ->booleanNode('additional_product_filters')
+                                    ->info('Indexes the product family and the attribute values as filter fields of the website index.')
+                                    ->defaultTrue()
```
Prokyonn: Default this to `false`: the bundle ships no catalogue page, so projects should opt in to the filter fields and to the drop-and-reindex every new number or date attribute needs.

src/Infrastructure/Sulu/Search/Visitor/WebsiteProductDetailsReindexProviderEnhancer.php
```diff
@@ -0,0 +1,383 @@
...
+use Sulu\Product\Domain\Model\ProductAttributeValue;
+use Sulu\Product\Domain\Model\ProductInterface;
+use Symfony\Contracts\Service\ResetInterface;
+
+/**
+ * Indexes the details tab: family and attribute values as filter fields, their text as content,
+ * and the details image as fallback. A variant also gets its parent's non-variant-specific values.
+ *
+ * Attribute definitions are loaded once per run and kept until the kernel resets the service;
+ * values are loaded per batch of the provider's query, so memory stays bounded by the batch size.
```
Prokyonn: This class does two jobs. External identifier, family name, short description and image belong in an enhancer that always runs; today `additional_product_filters: false` removes them from the site search too.

src/Infrastructure/Sulu/Search/Visitor/WebsiteProductDetailsReindexProviderEnhancer.php
```diff
@@ -0,0 +1,383 @@
...
+                        $display = $valueRow['text'];
+                    }
+                    break;
+            }
+
+            // "Widerstand: 5 Ω", so both label and value are searchable.
+            if (null !== $display) {
+                $label = $this->translate($attribute['labels'], $locale, $attribute['defaultLocale']) ?? $key;
+                $content[] = $label . ': ' . $display;
+            }
```
Prokyonn: I think the `Label: value` lines can go and option names and text values are enough in `content`: the label adds noise ("weight" matches every product) and the formatting duplicates `sulu_product_format_attribute_value`. What do you think?

src/Infrastructure/Sulu/Search/Schema/ProductSchemaLoader.php
```diff
@@ -0,0 +1,111 @@
...
+     */
+    private function loadNumericFields(): array
+    {
+        /** @var array<int, array{key: string}> $rows */
+        $rows = $this->entityManager->createQueryBuilder()
+            ->select('attribute.key AS key')
+            ->from(Attribute::class, 'attribute')
+            ->where('attribute.type IN (:types)')
+            ->setParameter('types', self::NUMERIC_TYPES)
+            ->orderBy('attribute.key', 'ASC')
```
Prokyonn: Build the filter fields from `isFilterable()` instead of the type, here and in `WebsiteProductDetailsReindexProviderEnhancer::enhanceDocument()`: an attribute with Filterable off, and every text attribute, still ends up as a filter field.

src/Infrastructure/Sulu/Search/Schema/ProductSchemaLoader.php
```diff
@@ -0,0 +1,111 @@
...
+        $index = $schema->indexes['website'] ?? null;
+        $product = $index?->fields[WebsiteProductDetailsReindexProviderEnhancer::FIELD] ?? null;
+        if (null === $index || !$product instanceof Field\ObjectField) {
+            return $schema;
+        }
+
+        $numericValues = $product->fields[WebsiteProductDetailsReindexProviderEnhancer::NUMERIC_VALUES_FIELD] ?? null;
+        if (!$numericValues instanceof Field\ObjectField) {
+            return $schema;
+        }
```
Prokyonn: If this loader adds the whole `product` field, `config/schemas/website.php`, the prepend block and `isAdditionalProductFiltersEnabled()` can go, and so can these checks for a field the loader always registers.

src/Infrastructure/Sulu/Search/Visitor/WebsiteProductDetailsReindexProviderEnhancer.php
```diff
@@ -0,0 +1,383 @@
...
+        return self::sanitize($attributeKey);
+    }
+
+    /**
+     * Reduces the key to word characters, the only ones search engines accept in field names.
+     */
+    private static function sanitize(string $key): string
+    {
+        return (string) \preg_replace('/[^A-Za-z0-9_]/', '_', $key);
+    }
```
Prokyonn: `sanitize()` allows a leading digit and the schema loader skips such keys, so a number attribute `3d_depth` is silently not filterable. If `sanitize()` also prefixes a leading digit, the skip in the loader can go.

src/Infrastructure/Sulu/Search/WebsiteProductReindexProvider.php
```diff
@@ -79,40 +81,22 @@ public function provide(ReindexConfig $reindexConfig): \Generator
...
 
                 if ('' === $data['title']) {
-                    $data['title'] = $product['title'];
+                    $data['title'] = (string) $row['title'];
+                }
+
+                // The content enhancer resets `content`, so the code is appended after the enhancers.
+                if (null !== $row['code'] && '' !== $row['code'] && \is_array($data['content'])) {
+                    $data['content'][] = $row['code'];
                 }
```
Prokyonn: If `WebsiteProductReindexContentEnhancer` appended to `content` instead of resetting it, the code could go into `createDocument()` and the `priority: -10` workaround would go too.

src/Infrastructure/Sulu/Search/WebsiteProductReindexProvider.php
```diff
@@ -123,48 +107,88 @@ public function provide(ReindexConfig $reindexConfig): \Generator
...
+    /**
+     * @param Product $row
+     * @param array<string, string[]> $webspaces keyed by "<productId>__<locale>"
+     *
+     * @return array<string, mixed>
+     */
+    private function createDocument(array $row, array $webspaces): array
+    {
+        // A variant owns its route, but its webspaces are the parent's.
+        $isVariant = ProductInterface::TYPE_VARIANT === $row['type'] && null !== $row['parentId'];
```
Prokyonn: `null !== $row['parentId']` is enough, only variants have a parent. Same in the enhancer.

src/Application/MessageHandler/CreateAttributeMessageHandler.php
```diff
@@ -44,6 +47,8 @@ public function __invoke(CreateAttributeMessage $message): AttributeInterface
         $this->attributeRepository->save($attribute);
         $this->attributeGroupRepository->save($group);
 
+        $this->domainEventCollector->collect(new AttributeCreatedEvent($attribute, $message->getLocale(), $message->getData()));
```
Prokyonn: `AttributeCreatedEvent` can go with the rebuild listener: it only adds an untranslated `sulu_activity.description.attributes.created` entry to the activity log, and no event covers changing or removing an attribute.

src/Infrastructure/Sulu/Search/Visitor/WebsiteProductDetailsReindexProviderEnhancer.php
```diff
@@ -0,0 +1,383 @@
...
+        ];
+
+        return $document;
+    }
+
+    /**
+     * Own values win; the parent's variant-specific values are not inherited.
+     *
+     * @return array<int, ValueRow> attribute id => row
+     */
```
Prokyonn: The docblocks here and on `productAttributeValues()`, `translate()` and `loadAttributeValues()` repeat the method name and can go.

tests/Functional/Infrastructure/Sulu/Search/WebsiteProductDetailsReindexProviderEnhancerTest.php
```diff
@@ -0,0 +1,370 @@
...
+        $this->assertIsArray($secondProduct);
+        $this->assertSame(['weight' => [3.0]], $secondProduct['attributes_numeric_values']);
+        $this->assertSame([], $secondProduct['attributes_text_values']);
+        $this->assertIsArray($second['content']);
+        $this->assertNotContains('Note: First note', $second['content']);
+    }
+
+    /**
+     * MySQL cuts GROUP_CONCAT at 1024 bytes by default; the values are loaded with it, so a product
+     * whose packed values exceed that must still be indexed completely.
```
Prokyonn: This test can go: the values are no longer loaded with `GROUP_CONCAT`, so it guards a limit the code no longer has.

tests/Unit/Infrastructure/Sulu/Search/WebsiteProductIndexListenerTest.php
```diff
@@ -33,48 +32,168 @@ class WebsiteProductIndexListenerTest extends TestCase
...
+        $listener->onProductChanged(new ProductRemovedEvent('parent-uuid', 'Gone', [
+            'locales' => ['en'],
+            'variantUuids' => ['variant-one-uuid', 'variant-two-uuid'],
+        ]));
     }
 
-    public function testOnProductChangedWithProductTranslationRemovedEvent(): void
+    /**
+     * A variant document holds its parent's url, so removing the parent's translation reindexes
+     * the whole family in that locale.
```
Prokyonn: A variant document has its own url, not its parent's. It is the parent's webspaces and shared values that make this reindex necessary.

tests/Functional/Infrastructure/Sulu/Search/ProductReindexProviderQueryCountTest.php
```diff
@@ -0,0 +1,238 @@
...
+            'name' => 'Query Count Family',
+            'attributes' => $attributes,
+        ]) ?: null);
+        $this->assertHttpStatusCode(201, $this->client->getResponse());
+        $this->familyId = $this->responseId();
+    }
+
+    /**
+     * A parent is an admin document and its variant a website document; the variant makes the
+     * website provider load the parent's webspaces, slug and shared values.
```
Prokyonn: The provider loads the parent's webspaces and shared values, not its slug.

Prokyonn (changes requested):

src/Infrastructure/Sulu/Search/Visitor/WebsiteProductDetailsReindexProviderEnhancer.php
```diff
@@ -0,0 +1,383 @@
...
+        $rows = (clone $queryBuilder)
+            ->select('product.uuid AS productId', 'IDENTITY(product.parent) AS parentId')
+            ->getQuery()
+            ->getArrayResult();
+
+        return \array_values(\array_unique(\array_filter(\array_merge(
+            \array_column($rows, 'productId'),
+            \array_column($rows, 'parentId'),
+        ))));
+    }
```
Prokyonn: Paging with `id > :lastId` instead of `setFirstResult()`, and reading the batch ids before the family joins are added, takes a full reindex of 15,000 products from 45 s to 6.5 s. Today both queries sort the whole live set on every batch.

PR #423, opened 2026-09-15
Update product resolving

martinlagler:
Related to sulu/sulu#9101 (merged), which sends the product type as the `productType` metadata option of the `product_attributes` field.

#### What's in this PR?

Variant attributes on products without variants, and website resolving that follows the variant routes from #419: a variant is a page of its own URL, `?variant=` is gone.

**Attributes in the admin**

- A family attribute with the "variant" toggle is now also held by a product without variants. `ProductAttributeScope` decides by product type: a product without variants holds shared and variant attributes, a product with variants only shared ones, a variant only variant ones.
- The `product_attributes` form, the details form schema and the required check in `ProductAttributesDataMapper` follow that rule. The form takes a `productType` metadata option, sent by the admin field.

**Website resolving**

- A variant page renders its parent's content tab: `ProductVariantDimensionContentEnhancer` resolves a variant with its parent's template, excerpt and SEO data. The route stays the variant's. `ProductParentContentLoader` loads the parent once per request for the enhancer and the resolver.
- The `product` namespace on a variant page:
  - `product`: the parent (`title`, master data, `attributes`, `associations`, `variants`)
  - `product.currentVariant`: the variant (`title`, `url`, `code`, ..., its own `attributes`)
  - `product.variants`: every published variant with `title`, `url`, `code`, `status`, `position`, plus the fields from `sulu_product.variants.fields`
- A product without variants resolves as itself with `product.url`; a product with variants has no `url`.
- A variant page is cache-tagged with its parent and every listed variant, so publishing the parent or a sibling clears it.
- Attributes are no longer merged by the resolver. `sulu_product_merge_attributes(product.attributes, product.currentVariant.attributes)` combines them in the template; `sulu_product_format_attribute_value` formats a single value.

**Removed**

- The website `ProductController`, the `variant_query_parameter` config, the parent-slug URL of variants in `ProductResourceLoader` and `ProductRepositoryInterface::findSlugsBy()`.

**Also**

- `ProductParentMapper` keeps the position when the variant form sends an empty one (it failed with "Invalid attribute value provided.").

#### BC Breaks

- `selectedVariant` is replaced by `product.currentVariant`; `product.variants` entries are flat product fields instead of resolved references (`variant.title` instead of `variant.content.title`, `variant.code` instead of `variant.product.code`).
- `product.attributes` holds a product's own values only.
- Product templates using `Sulu\Product\UserInterface\Controller\Website\ProductController` switch to `Sulu\Content\UserInterface\Controller\Website\ContentController`.
- `sulu_product.variant_query_parameter` is removed; `sulu_product.variants.fields` is new.
- `title`, `url` and `currentVariant` are reserved `details/<field>` names.

The README describes the new namespace.


src/Application/Attribute/ProductVariantAttributesMerger.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public function merge(iterable $productAttributes, iterable $variantAttributes): array
+    {
+        $merged = [];
+
+        foreach ([$productAttributes, $variantAttributes] as $attributes) {
+            foreach ($attributes as $value) {
+                $merged[$value->getAttribute()->getKey()] = $value;
+            }
+        }
```
Prokyonn, line 38: if those 4 lines is really all we need to merge it, I'm not sure if this is worth as a separate service. The Twig extension test already covers the merge, so the service, the alias and its test could go too.

src/Domain/Model/ProductAttributeScope.php
```diff
@@ -0,0 +1,31 @@
...
+namespace Sulu\Product\Domain\Model;
+
+/**
+ * Which family attributes a product holds by its type: a variant only variant attributes, a product
+ * with variants only shared ones, a product without variants both, since nothing else holds its
+ * variant attributes.
+ */
+final class ProductAttributeScope
+{
+    public static function holds(string $productType, ProductFamilyAttributeInterface $familyAttribute): bool
```
Prokyonn, line 23: is this really a domain model? Where is this used so far?

Prokyonn: also this would be more about the attributefamily scope right?

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -41,11 +50,20 @@ class ProductResolver implements ResolverInterface
 
     private const ASSOCIATIONS_FIELD_PREFIX = 'associations/';
 
+    private const VARIANT_FIELDS = ['title', 'url', 'code', 'status', 'position'];
```
Prokyonn: default variant fields?

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -41,11 +50,20 @@ class ProductResolver implements ResolverInterface
...
+
+    /**
+     * @param list<string> $variantFields product fields each entry of `variants` carries in addition to title, url, code, status and position
```
Prokyonn: no description, this comment will just be out of sync with the actual field

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -279,7 +333,7 @@ private function resolveDetailsData(
      *
      * @return array<string, ProductAttributeValueInterface>
      */
-    private function resolveAttributes(ProductDimensionContentInterface $dimensionContent): array
+    private function keyAttributes(ProductDimensionContentInterface $dimensionContent): array
```
Prokyonn, line 334: rename to something like getAttributesMapByKey or something similar that is more informative

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -182,13 +229,20 @@ private function resolveDetails(
         ?array $requested,
     ): array {
         $fixed = [
```
Prokyonn: more informal variable name, fixed sounds weird

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -321,49 +375,49 @@ private function resolveAssociations(ProductDimensionContentInterface $dimension
     }
 
     /**
-     * No property projection, so a variant reads like the page it sits on (`variant.product.image`).
-     * A variant is not itself `product_with_variants`, so this does not nest.
+     * Each variant as flat product fields (`title`, `url`, `code`, `status`, `position` plus the
```
Prokyonn, line 376: less details in the comment, I'm not even sure if we need it, but naming every field will only go out of sync with the actual implementation

src/Infrastructure/Symfony/HttpKernel/SuluProductBundle.php
```diff
@@ -238,10 +239,15 @@ public function configure(DefinitionConfigurator $definition): void
...
                     ->end()
                 ->end()
-                ->scalarNode('variant_query_parameter')
-                    ->info('Query parameter a variant URL carries, e.g. /product/xy?variant=XY-2.')
-                    ->defaultValue('variant')
-                    ->cannotBeEmpty()
+                ->arrayNode('variants')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->arrayNode('fields')
```
Prokyonn: this should be a property map, also we should probably prepend the default config, so that we dont need it in the resolver. Not sure but I think the config will be merged with the one from the project, so it would be easier to add more properties

That map could then be what we pass as `properties` to the variant resolvables (see the comment on `resolveVariants()`), and the defaults could be the same set `getDefaultProperties()` uses for a product selection, so a variant entry and a selection tile carry the same fields.

src/Infrastructure/Symfony/Twig/ProductAttributeTwigExtension.php
```diff
@@ -30,22 +33,42 @@ class ProductAttributeTwigExtension extends AbstractExtension
...
+        return $this->attributesMerger->merge($productAttributes, $variantAttributes);
+    }
+
+    /**
+     * @param iterable<ProductAttributeValueInterface> $productAttributes keys are ignored; re-keyed by attribute key
      *
      * @return list<array{key: string, label: string, attributes: array<string, ResolvedAttribute>}>
      */
-    public function groupAttributes(iterable $attributes, ?string $locale = null): array
+    public function groupAttributes(iterable $productAttributes, ?string $locale = null): array
```
Prokyonn, line 71:
```suggestion
    public function groupAttributes(iterable $productAttributeValues, ?string $locale = null): array
```

src/Infrastructure/Symfony/Twig/ProductAttributeTwigExtension.php
```diff
@@ -103,17 +126,26 @@ public function groupAttributes(iterable $attributes, ?string $locale = null): a
...
-    ): ?string {
+    /**
+     * The value as displayed: option name, text or number with its display format, date per locale.
+     * Without a locale the request's; null for an empty value or without any locale.
+     */
+    public function formatValue(ProductAttributeValueInterface $productAttributeValue, ?string $locale = null): ?string
+    {
+        $locale ??= $this->requestAnalyzer->getCurrentLocalization()?->getLocale();
+
+        if (null === $locale) {
```
Prokyonn, line 137: locale is only required for options and dates (`formatDate()` passes it to `IntlDateFormatter`), text and number values could still be formatted without one, right?

src/Infrastructure/Sulu/Content/ContentEnhancer/ProductVariantDimensionContentEnhancer.php
```diff
@@ -0,0 +1,68 @@
...
+        if (null === $parentContent) {
+            return $dimensionContent;
+        }
+
+        $templateKey = $parentContent->getTemplateKey();
+        if (null !== $templateKey) {
+            $dimensionContent->setTemplateKey($templateKey);
+        }
+
+        $dimensionContent->setTemplateData($parentContent->getTemplateData());
```
Prokyonn, line 53: Like core's `PageLinkDimensionContentEnhancer`, I'd keep the variant's own title here: `[...$parentContent->getTemplateData(), 'title' => $dimensionContent->getTitle()]`. The enhancer also runs for selections and teasers, so any template or excerpt field a tile picks comes from the parent, and all variant pages share the parent's `content.title`. That also makes the `isMerged()` check in `ProductDimensionContent::setTemplateData()` unnecessary. Is the parent's excerpt on a variant teaser intended?

src/Infrastructure/Sulu/Content/ContentEnhancer/ProductVariantDimensionContentEnhancer.php
```diff
@@ -0,0 +1,68 @@
...
+class ProductVariantDimensionContentEnhancer implements DimensionContentEnhancerInterface
+{
+    public function __construct(
+        private readonly ProductParentContentLoader $parentContentLoader,
+    ) {
+    }
+
+    public function enhance(DimensionContentInterface $dimensionContent): DimensionContentInterface
+    {
+        if (!$dimensionContent instanceof ProductDimensionContentInterface || !$dimensionContent->isMerged()) {
```
Prokyonn, line 39: Enhancers only get merged content (the content resolver and the teaser providers aggregate first), so I think the `isMerged()` check and its test can go.

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -65,13 +83,39 @@ public function resolve(DimensionContentInterface $dimensionContent, ?array $pro
 
         $requested = null === $properties
             ? null
-            : $this->filterProperties(\array_merge($this->getDefaultProperties(), $properties));
+            : [...$this->filterProperties($this->getDefaultProperties()), ...$this->filterProperties($properties)];
+
+        // A reference, such as a product selection, resolves a variant as itself.
+        $parentContent = null === $requested ? $this->parentContentLoader->load($dimensionContent) : null;
```
Prokyonn, line 87: Could the parent be a resolvable instead, as `product.parent` via `ContentView::createResolvable()`? Core then loads, tags and references it, the parent's own resolver run brings the `variants`, and `product` stays the variant everywhere. Right now the swap depends on `$properties === null`, which `ProductReferenceRefresher` also passes, so a variant's stored references include the parent's associations and all its siblings. The `ProductParentContentLoader` and its request cache would also go; the enhancer can load the parent inline like `PageLinkDimensionContentEnhancer`.

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -321,49 +375,49 @@ private function resolveAssociations(ProductDimensionContentInterface $dimension
...
     /**
-     * No property projection, so a variant reads like the page it sits on (`variant.product.image`).
-     * A variant is not itself `product_with_variants`, so this does not nest.
+     * Each variant as flat product fields (`title`, `url`, `code`, `status`, `position` plus the
+     * configured fields), the shape of `currentVariant`. Built here rather than as references: a projected
+     * reference moves its fields under `content`. A variant is not itself `product_with_variants`, so
+     * this does not nest.
      */
-    private function resolveVariants(ProductDimensionContentInterface $dimensionContent, string $locale): ?ContentView
+    private function resolveVariants(ProductInterface $product, string $locale): ?ContentView
```
Prokyonn, line 381: I'd go back to `createResolvablesWithReferences()` with the variant fields as `properties`, so batching, cache tags and references come from core. The `content` nesting of projected references is a gap left by sulu/sulu#9046: `recursivelyMapProperties()` writes to `content` instead of the resolver's output path. I'd rather fix that in core than work around it here.

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -182,13 +229,20 @@ private function resolveDetails(
...
+            'title' => ContentView::create($dimensionContent->getTitle(), []),
+            'url' => ContentView::create($dimensionContent->getRoute()?->getSlug(), []),
             'code' => ContentView::create($dimensionContent->getCode(), []),
             'externalIdentifier' => ContentView::create($dimensionContent->getExternalIdentifier(), []),
             'productFamily' => $this->resolveProductFamily($dimensionContent, $locale),
             'status' => ContentView::create($dimensionContent->getStatus(), []),
             'position' => ContentView::create($dimensionContent->getResource()->getPosition(), []),
         ];
 
+        // A product with variants owns no route, it is reached through its variants.
```
Prokyonn, line 239: A product with variants never has a route, so `url` is null anyway. I'd drop the `unset`; a variant without a route also gets `url: null`.

src/Infrastructure/Sulu/Admin/ProductAttributesFormMetadataVisitor.php
```diff
@@ -88,6 +90,24 @@ public function visitFormMetadata(FormMetadata $formMetadata, string $locale, ar
         $formMetadata->setItems($items);
     }
 
+    /**
+     * Without a type only shared attributes, as for a product with variants.
+     *
+     * @param array<string, mixed> $metadataOptions
+     */
+    private function resolveProductType(array $metadataOptions): string
```
Prokyonn, line 98: Two options for the same thing now: `variant` from the XML and `productType` from the admin field. Could the variant form send `productType: variant` so there is one option (small change in sulu/sulu's `ProductAttributes.js`)? Currently `productType: variant` falls back to `product_with_variants`.

README.md
```diff
@@ -34,6 +34,44 @@ carries an address of its own. A product with variants owns none, it is reached
...
+A product without variants resolves as itself, with `product.url` and without `variants` or
+`currentVariant`. A variant resolved as a reference (a product selection) resolves as itself.
+
+To show a variant's attributes together with its product's, merge them in the template; the
+variant's value wins on the same attribute:
+
+```twig
+{% set attributes = sulu_product_merge_attributes(product.attributes, product.currentVariant.attributes|default({})) %}
+{% set groups = attributes|sulu_product_attribute_groups %}
+```
```
Prokyonn, line 74: Could we add `sulu_product_format_attribute_value` here too? It's meant for projects, e.g. showing a variant's option value in a variant switcher, but right now only the code knows it exists:

```twig
{{ attributeValue|sulu_product_format_attribute_value }}
```

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
```diff
@@ -16,21 +16,30 @@
...
 use Sulu\Product\Domain\Model\ProductInterface;
 use Sulu\Product\Domain\Repository\ProductRepositoryInterface;
-use Sulu\Product\Infrastructure\Sulu\Content\ResourceLoader\ProductResourceLoader;
+use Sulu\Product\Infrastructure\Sulu\Content\ProductParentContentLoader;
 
 /**
  * Assembles the root-level `product` namespace. A reference passes `$properties` and gets the
  * always-on set plus what it asked for under `product.`, so a listing builds no full payload.
  *
+ * A variant resolved in full, as the page of its route, resolves its parent as `product` and itself
```
Prokyonn, line 38: I'd drop this paragraph, it repeats the enhancer docblock and the README.

src/Infrastructure/Sulu/Content/ContentEnhancer/ProductVariantDimensionContentEnhancer.php
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Sulu\Product\Infrastructure\Sulu\Content\ContentEnhancer;
+
+use Sulu\Content\Application\ContentEnhancer\DimensionContentEnhancerInterface;
+use Sulu\Content\Domain\Model\DimensionContentInterface;
+use Sulu\Product\Domain\Model\ProductDimensionContentInterface;
+use Sulu\Product\Infrastructure\Sulu\Content\ProductParentContentLoader;
+
+/**
+ * A variant owns its route and product fields but no content: it is resolved with its parent's
```
Prokyonn, line 22: One line is enough here.

src/Infrastructure/Sulu/Admin/ProductAttributesSchemaFormMetadataVisitor.php
```diff
@@ -60,6 +62,8 @@ public function visitFormMetadata(FormMetadata $formMetadata, string $locale, ar
     }
 
     /**
+     * One branch per family and product type, so the attributes follow the type select live.
```
Prokyonn, line 65: Repeats the class docblock.

Prokyonn (changes requested):

src/Domain/Model/ProductAttributeScope.php
martinlagler: We create a `isAvailable($productType)` function in  ProductFamilyAttribute

src/Infrastructure/Sulu/Content/Resolver/ProductResolver.php
martinlagler: will be fixed in a follow up

PR #426, opened 2026-09-16
Add is filterable to product attribute

martinlagler:
Adds a `filterable` checkbox to product family attributes, stored on `ProductFamilyAttribute` and returned by the product family API. It defaults to `false`.

Projects need a migration:

```sql
ALTER TABLE pr_product_family_attributes ADD filterable TINYINT DEFAULT 0 NOT NULL;
```

Admin UI: sulu/sulu#9108


tests/Functional/Integration/ProductFamilyControllerTest.php
```diff
@@ -208,7 +208,7 @@ public function testPostWithFamilyAttributesRoundTrip(): void
                 'name' => 'Apparel',
                 'description' => null,
                 'attributes' => [
-                    ['id' => $attributeUuid, 'required' => true, 'variantSpecific' => false],
+                    ['id' => $attributeUuid, 'required' => true, 'variantSpecific' => false, 'filterable' => false],
```
Prokyonn, line 211: Sending `'filterable' => true` here and in the assertion below would cover the new Doctrine mapping. With `false` the round trip also passes without the mapping, because `false` is the property default.

tests/Functional/Integration/ProductVariantControllerTest.php
```diff
@@ -69,6 +69,7 @@ private function createProductFamily(array $attributes = []): string
                 'id' => $attribute->getUuid(),
                 'required' => $entry['required'] ?? false,
                 'variantSpecific' => $entry['variantSpecific'] ?? false,
+                'filterable' => false,
```
Prokyonn, line 72: `createProductFamily()` always sends `false` here and never reads `$entry['filterable']`, which isn't in its `@param` shape either. So the `'filterable' => false` added at the call sites (lines 235, 403, 404, 850, 851, 907, 936, 962, 1010, 1331, 1379) has no effect and can be reverted.

martinlagler:
Filterable should be on Attribute and not on ProductFamilyAttribute

PR #427, opened 2026-09-16
Add publish/unpublish toolbar action for variants

martinlagler:
Requires sulu/sulu#9111, which adds the `sulu_admin.publishing` list toolbar action. Merge that first: without it the admin fails on the unknown toolbar action.

#### What's in this PR?

Variants are published on their own, from the variants tab of their product:

- **Toolbar:** the variants list gets the `sulu_admin.publishing` toolbar action (live permission only), plus `published`, `publishedState` and `ghostLocale` in the list.
- **Endpoint:** `POST /admin/api/products/{parentId}/variants/{id}?action=publish|unpublish` applies the transition and answers `409` if it is not available, including publishing a variant whose product is not published in that locale.
- **Publish state follows the product:** unpublishing a product or removing its translation unpublishes its variants in that locale; publishing it leaves them as they are. Saving a variant no longer marks its product as changed (`VariantWorkflowCascader` and `VariantParentPublishStateUpdater` are removed).
- **Website:** a variant whose product is not live in that locale answers 404.
- **Loader fix:** `ProductParentContentLoader` refreshes a managed parent before loading another locale or stage, so a variant no longer resolves against the parent content loaded earlier.

#### Why?

Variants need their own publish state, but a variant renders its product's content, so it can only be live while its product is. Moving the toolbar action into sulu/sulu keeps the bundle free of admin JavaScript, so projects don't need to add it to their admin build.


README.md
```diff
@@ -34,6 +34,23 @@ carries an address of its own. A product with variants owns none, it is reached
...
+A variant is published on its own, from the variants tab of its product: select variants in the
+list and use "Publishing" in the toolbar, which needs the live permission. Publish skips variants
+without content in the current locale, unpublish only acts on published ones. The list marks
+unpublished variants and variants shown in a fallback locale, the same way the article list does.
+
+A product with variants and its variants never change each other's publish state: publishing or
+unpublishing the product leaves its variants as they are, and saving a variant does not mark its
+product as changed. The website lists and routes only published variants. A published variant
+whose product is not published has no product content to inherit, so publish the product as
+well.
```
Prokyonn: Unpublishing a product with variants no longer takes anything off the website. I published a variant, then set its product to unpublished from the Details tab. The variant URL still answers 200 and shows only the variant's own title, without the product's title, description or other content. A product with variants has no URL of its own, so unpublishing it now only breaks its variant pages.

Could `ProductRouteDefaultsProvider::getDefaults()` answer 404 for a variant whose product has no live content in that locale? Variants would keep their own publish state, and this sentence could go.

src/Infrastructure/Sulu/Content/ProductParentContentLoader.php
```diff
@@ -65,6 +70,19 @@ public function reset(): void
...
+    /**
+     * A managed parent keeps the dimension contents an earlier query loaded, and a fetch join does not
+     * replace them, so another locale or stage would resolve against the wrong rows.
+     */
+    private function dropLoadedDimensionContents(ProductInterface $parent): void
+    {
+        $dimensionContents = $parent->getDimensionContents();
+
+        if ($dimensionContents instanceof PersistentCollection && $dimensionContents->isInitialized()) {
+            $this->entityManager->refresh($parent);
```
Prokyonn, line 82: No functional test depends on this refresh. With this line removed, `ProductVariantControllerTest` (39 tests) and `ProductResolverVariantsTest` (2 tests) still pass; only the unit test that expects `refresh()` to be called fails. Could you add a functional test for the case from the description, a variant resolved against parent content loaded earlier in the same request, that fails without the refresh?

src/Infrastructure/Sulu/Route/ProductRouteDefaultsProvider.php
```diff
@@ -78,6 +80,11 @@ public function getDefaults(Route $route): array
             throw new NotFoundHttpException(\sprintf('No product found for id "%s" and locale "%s"', $id, $locale), $exception);
         }
 
+        // a variant renders its product's content, so it is only reachable while that product is live
+        if (null !== $product->getParent() && null === $this->parentContentLoader->load($dimensionContent)) {
+            throw new NotFoundHttpException(\sprintf('No published parent product found for variant id "%s" and locale "%s"', $id, $locale));
+        }
```
Prokyonn, line 86: The variant page answers 404 now, thanks. The sitemap and the website search still list a variant whose product is unpublished, so both link to that 404: after unpublishing the product, `/sitemaps/products-1.xml` still contains the variant URL, and searching for the variant's title still shows it.

`ProductsSitemapProvider::getProducts()` only checks the variant's own live content, and `WebsiteProductReindexProvider::loadBatch()` indexes any live content. Unpublishing the product also reindexes only the product, not its variants. Could the sitemap and the search index skip a variant whose product has no live content in that locale, and could publishing or unpublishing a product reindex its variants?

martinlagler: Solved by preventing the state instead of filtering it: a variant renders its product's content, so it is now only live together with its product.

- Unpublishing a product or removing its translation unpublishes its variants in that locale.
- Publishing a variant answers `409` while its product is not published in that locale. The list shows the reason for each variant it could not publish.
- Publishing the product leaves its variants as they are.

So the sitemap and the website search can no longer list a variant whose product is offline, and the variants that go offline are reindexed through their own publish events. The 404 in `ProductRouteDefaultsProvider` stays as the guard for the page itself.

Prokyonn (approved):

PR #428, opened 2026-09-17
Add filterable to attribute

martinlagler:
Adds a `filterable` flag to attributes (toggle in the admin form). It is only kept for `number`, `date` and `options` attributes, because free text has no discrete values to filter by.

Also lets date attributes use the display format as an ICU pattern (e.g. `MMMM yyyy`), still rendered in the requested locale.


src/Infrastructure/Symfony/Twig/ProductAttributeTwigExtension.php
```diff
@@ -179,14 +180,26 @@ private function applyDisplayFormat(AttributeInterface $attribute, string|float|
...
-        $formatter = new \IntlDateFormatter($locale, \IntlDateFormatter::MEDIUM, \IntlDateFormatter::NONE);
+        $format = $attribute->getConfig()['displayFormat'] ?? null;
+
+        $formatter = new \IntlDateFormatter(
+            $locale,
+            \IntlDateFormatter::MEDIUM,
+            \IntlDateFormatter::NONE,
+            null,
+            null,
+            \is_string($format) && '' !== $format ? $format : null,
```
Prokyonn, line 202: A word in the pattern breaks the output, and nobody sees an error. I saved `Released MMMM yyyy` on a date attribute: the admin accepts it, and the date disappears from the product page. ICU returns an empty string for an unknown letter like `R`, and `groupAttributes()` drops empty values. Other words print garbage instead: `%value%` renders as `%GMT+0AM`. Editors will try this, because the same field takes free text around `%value%` for text and number attributes.

Could saving a date attribute reject a pattern that formats a sample date to an empty string? The info text could also say that literal words go in single quotes, e.g. `'Released' MMMM yyyy`.

src/Infrastructure/Symfony/Twig/ProductAttributeTwigExtension.php
```diff
@@ -179,14 +180,26 @@ private function applyDisplayFormat(AttributeInterface $attribute, string|float|
...
         }
 
-        $formatter = new \IntlDateFormatter($locale, \IntlDateFormatter::MEDIUM, \IntlDateFormatter::NONE);
+        $format = $attribute->getConfig()['displayFormat'] ?? null;
+
+        $formatter = new \IntlDateFormatter(
+            $locale,
+            \IntlDateFormatter::MEDIUM,
+            \IntlDateFormatter::NONE,
+            null,
```
Prokyonn: `DateAttributeType` stores a date as midnight UTC, but this formatter uses the server's default timezone. West of UTC every date shows a day early, and with a month pattern the 1st of a month shows the previous month: with `America/New_York`, `2024-03-01` renders as "February 2024". The medium format already showed "Feb 29, 2024" before this PR, but this call is being rewritten anyway: passing `'UTC'` as the timezone here fixes both.

tests/Functional/Integration/AttributeControllerTest.php
```diff
@@ -629,6 +629,44 @@ public function testPersistsFormatInConfig(): void
         $this->assertSame('> %value% GΩ', $config['displayFormat']);
     }
 
+    public function testFilterableRoundTripsAndIsRejectedForText(): void
```
Prokyonn, line 632: This test sends `filterable` only when creating an attribute. Switching it on an existing attribute goes through the update path, and no test covers that: if the update branch in `AttributeMapper` never stored `true`, every test here and in `AttributeMapperTest` would still pass. A PUT that switches it on for an existing number attribute, and one that tries it on an existing text attribute, would cover it.

Prokyonn (changes requested):
**Nothing reads `isFilterable()` yet, and #420 disagrees with the type rule.** #420 turns every number and date attribute into a filter field by its type (`loadNumericFields()`), and it indexes text values as filterable facets too (its test expects `note:Gold plated`). Once both are merged, the switch changes nothing, and text is filterable in the index while this PR says it cannot be. Which rule should hold? If text stays excluded, #420 should build its fields from the flag instead of the type; if #420 is right about text, the type restriction here goes.

**The date display format is a second feature.** The squash commit will only say "Add filterable to attribute", so the ICU pattern change will be hard to find later. I'd split it out, or at least name it in the title.

Details inline.

Prokyonn (approved):
Approving on the answer in the description: the flag is the rule, so #420 has to build its filter fields from `isFilterable()` instead of the attribute type. Until that lands, the switch has no effect.

PR #429, opened 2026-09-17
Link product attribute values to their attribute option

martinlagler:
Replaces the stored `attributeOptionKey` on `ProductAttributeValue` with the existing relation to `AttributeOption`, and removes the unused `uuid` from `AttributeOption`.

- The `ProductAttributeValue` constructor accepts an optional `AttributeOptionInterface`. `setAttributeOptionKey()` is removed, and `getAttributeOptionKey()` now reads the key from the option.
- `OptionsAttributeType::writeValue()` resolves the submitted key via `Attribute::getOption()` and rejects unknown keys.

**BC break:** the columns `pr_product_attribute_values.attributeOptionKey` and `pr_attribute_options.uuid` are dropped. Fill `idAttributeOptions` from the stored key before dropping the column, otherwise option values are lost.


src/Domain/Model/AttributeOption.php
```diff
@@ -21,8 +21,6 @@ class AttributeOption implements AttributeOptionInterface
 {
     protected int $id;
 
-    protected ?string $uuid = null;
```
Prokyonn, line 24: #1 Removing `AttributeOption::$uuid` collides with #422, which makes that uuid the primary key of `pr_attribute_options` and renames the join column on the value to `attributeOptionUuid`. With both merged, one change drops the column and the next adds it back, and the note about `pr_attribute_options.uuid` in this description stops being true. Could the uuid removal move out of this PR, or is the option part of #422 off the table? It is a separable change from linking the values either way.

config/doctrine/Product/ProductAttributeValue.orm.xml
```diff
@@ -8,7 +8,6 @@
         </id>
 
         <field name="attributeKey" column="attributeKey" type="string" length="255" nullable="false"/>
-        <field name="attributeOptionKey" column="attributeOptionKey" type="string" length="255" nullable="true"/>
```
Prokyonn, line 11: #2 #420 still reads this field: `WebsiteProductDetailsReindexProviderEnhancer` selects `value.attributeOptionKey AS optionKey`, which stops resolving once the column and its mapping are gone. Whichever of the two merges second needs that query joined through `value.attributeOption`.

Prokyonn (commented):
Reviewed at ac653af9 against 3.0 (merge base 1bb751c). The relation swap itself looks right: no reader of the dropped column is left, `writeValue()` rejects an unknown key, and `ProductController` already turns the `InvalidArgumentException` into a 400. I ran the flow in a running app: selecting an option, saving and reloading keeps it, publishing renders the option's translated name on the website, and deleting an option a published product uses breaks neither the admin form nor the page.

The red jobs are not from this PR. Only `ProductResolverAssociationsTest::testPublishedTargetResolvesDeclaredProperties` fails, the same way on #428 and on an earlier push of #427, since sulu/sulu#9102 made the `product` output flat. #427 rewrites that assertion.

Four comments inline: #1 the uuid removal against #422, #2 the query in #420, #3 a missing functional test, #4 what a renamed option key does to existing values.

tests/Unit/Application/AttributeType/OptionsAttributeTypeTest.php
```diff
@@ -40,29 +41,39 @@ public function testKeyAndFormKey(): void
...
-        self::assertNull($value->getAttributeOptionKey());
+        self::assertNull($value->getAttributeOption());
 
         $type->writeValue($value, 'green');
         $type->writeValue($value, '');
-        self::assertNull($value->getAttributeOptionKey());
+        self::assertNull($value->getAttributeOption());
+    }
+
+    public function testWriteUnknownOptionKeyThrows(): void
```
Prokyonn, line 70: #3 The new behaviour is only covered in memory. No functional test writes an options value through the API and reads it back through the schema: the controller tests post number and text values, and nothing under `tests/Functional` touches `attributeOption`. Since the relation is now the only storage of the value, a broken mapping would pass here. A controller test that posts an option key, clears the entity manager, asserts the key on the draft and, after `?action=publish`, on live would cover it.

src/Domain/Model/ProductAttributeValue.php
```diff
@@ -55,14 +55,7 @@ public function getAttributeKey(): string
...
     {
-        return $this->attributeOptionKey;
-    }
-
-    public function setAttributeOptionKey(?string $key): self
-    {
-        $this->attributeOptionKey = $key;
-
-        return $this;
+        return $this->attributeOption?->getKey();
```
Prokyonn, line 58: #4 With the key read from the relation, an option's lifecycle now decides the value's. `AttributeMapper::mapOptions()` matches options by key, so renaming an option's key removes the old option and creates a new one, and the `ON DELETE SET NULL` on `idAttributeOptions` clears that value on every product that used it, published pages included and without a republish. Deleting an option does the same. Before this PR the stored key survived both. If a rename should keep the values, the mapper would have to update the existing option's key instead of recreating it; if the clearing is wanted, the description is the place to say so.

martinlagler: The question is should it be possible to change the an option key ? @Prokyonn @chirimoya

chirimoya: I don't know the data model of the Option, but an own ID would make sense to reverence the Option instead of the key as long as the key can be changed.

src/Domain/Model/AttributeOption.php
martinlagler: #422 will be merged after this one. Nothing reads the option uuid any more, so it can go here, and #422 adds it back as the primary key when it rebases. I'll update the description note about `pr_attribute_options.uuid`.

config/doctrine/Product/ProductAttributeValue.orm.xml
martinlagler: #420 will be merged after this one and rebases onto it. Its query will then select the key through `value.attributeOption`.

config/doctrine/Product/AttributeOption.orm.xml
```diff
@@ -30,10 +29,5 @@
                 <cascade-persist/>
             </cascade>
         </one-to-many>
-
-        <unique-constraints>
```
Prokyonn, line 34: #5 Small one for the description: this also drops the `UNIQ_pr_attr_opt_key` index, which the BC note doesn't mention yet. Could you add it together with the uuid update from #1?

src/Domain/Exception/AttributeOptionKeyNotUniqueException.php
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Product\Domain\Exception;
+
+class AttributeOptionKeyNotUniqueException extends \Exception
```
Prokyonn, line 16: #6 PHP Coverage is red: this exception's constructor and `getOptionKey()`, and the 409 for duplicate keys on create (`AttributeController.php:128-129`), have no test yet.

Prokyonn (changes requested):
Re-review at 09d76f9 (the same change, rebased onto the current 3.0).

#1 and #2 are fine with the merge order you described. #3 and #4 are fixed: in a running app, renaming an option's key keeps the value on products, a duplicate key shows a clear error, and duplicating an option block works. The tests are green; only the coverage job is red (#6). Two small points inline.

PR #430, opened 2026-09-17
Resolve variants through core resolvables

martinlagler:
Follow up to #423 and #425, uses sulu/sulu#9102.

#### What's in this PR?

- `product.variants` is resolved by core: `ContentView::createResolvablesWithReferences()` over the published variant uuids, with `sulu_product.variants.properties` as properties. Core loads, cache tags and references the variants, and sulu/sulu#9102 keeps the entries flat (`variant.title`). The hand-built variant loop is gone.
- The variant page swap (`product` = parent, `product.currentVariant` = variant) only happens when the variant enhancer loaded the parent. The reference index resolves without enhancers, so a variant no longer stores its parent's associations and all its siblings as references.

#### BC Breaks

- Referenced products below `product` are flat since sulu/sulu#9102, e.g. `product.associations.suitable[0].title` instead of `...suitable[0].content.title`.


Prokyonn:
The variants part looks right to me. `createResolvablesWithReferences()` with `sulu_product.variants.properties` is what was parked in #423 and #425, and I checked it in a running app on 653bd26: entries are flat and in position order, drafts stay out, the page's `X-Cache-Tags` carries the parent and every listed variant, and the reference index no longer stores a variant's parent associations and siblings.

My question is about the other half. The swap stays, and `getLoaded()` plus the `WeakMap` are there to tell a page render apart from every other resolve. That is the third mechanism for that same question, after the enhancer and after `$properties === null`, and the reason seems to be that the page needs the parent's content bag while the resolve root is the variant.

The `product.parent` idea from the #423 review does not get around it either: `ContentViewDataNormalizer::replaceNestedContentViewsAtPath()` replaces a resolved entity below a flat path by its `content` bag, so `product.parent` would carry the parent's template fields but not its `attributes`, `associations` or `variants`.

What feels like the best solution to me is to let routing decide what a variant URL renders. `ProductRouteDefaultsProvider` resolves the parent as `object` and carries the variant along as a second default, and `ProductResolver` adds `product.currentVariant` from it and tags the variant. The preview defaults provider does the same and keeps mapping the form data onto the variant. Then `ProductVariantDimensionContentEnhancer`, `ProductParentContentLoader`, the swap and the `WeakMap` all go, and nothing has to work out who is asking.

The one gap is `localizations`: core builds them from the resolved content, and a variant page is exactly the case where the route and the content are different resources. The TODO above `resolveSuluLocalizations()` and sulu/sulu#8175 already point there, so it would fit as a small core change.

Two decisions come with it: a variant whose parent has no published content in that locale becomes a 404, and variant teasers need `ProductTeaserProvider` to load the parent itself instead of leaning on the enhancer.

Nothing else from me on the diff. Would you rather merge this as it is and drop the swap machinery in that follow up, or fold it in here? What do you think of the direction?

Prokyonn (approved):

martinlagler:
Will be fixed in follow up PR.

PR #431, opened 2026-09-17
Add image to product family

martinlagler:
Based on #430, which needs to be merged first.

#### What's in this PR?

A product family gets an optional image (`single_media_selection`), shared by all locales.

- Admin API reads and writes `image: {id}`.
- On the website the image resolves as media at `product.productFamily.image`.

Projects need a migration:

```sql
ALTER TABLE pr_product_families ADD image_id INT DEFAULT NULL;
ALTER TABLE pr_product_families ADD CONSTRAINT FK_5CBB46C03DA5256D FOREIGN KEY (image_id) REFERENCES me_media (id) ON DELETE SET NULL;
CREATE INDEX IDX_5CBB46C03DA5256D ON pr_product_families (image_id);
```


Prokyonn (approved):
Works well in the browser. Picking, changing and clearing the image in the family form all save correctly. `product.productFamily.image` renders as a media on the product page, and also for an associated product that declares `productFamily`.

One thing to consider, maybe as a follow-up: the family image is written into the reference index once per product (`product.productFamily.image`), but saving a family doesn't refresh those rows or the product pages' cache. After switching a family from image A to image B, A's References tab still lists the product and B lists nothing. A cached product page keeps showing A until it expires or the product is published again. The family name already has the same cache gap. Tagging product pages with the family, invalidating that tag on family save and refreshing the products' references would cover both.

Could you also rebase onto 3.0? The branch still has the pre-squash commit of #430, so the PR shows 26 files instead of 18.

martinlagler:
Thanks for testing! Agreed on both points, but I'd keep them out of this PR. The cache gap already exists for the family name, so it deserves its own fix covering name, image and attributes together: tag product pages with the family and invalidate that tag on family save. While looking into it I also noticed the image shouldn't be recorded as a reference of every product in the first place, since it belongs to the family. Dropping those per-product rows will be part of the follow-up too.

I'll rebase onto 3.0.

PR #433, opened 2026-09-19
Register a workflow transition security context provider for products

### LINKED ISSUE #432: Register a workflow transition security context provider for products
### Summary

Since sulu/sulu#9100 ("Create, settle and authorize transition requests", merged on 3.1 on 2026-09-17), a workflow transition is authorized through `WorkflowTransitionRequestSecurityContextResolver`, which looks up a service tagged `sulu_content.workflow_transition_request_security_context_provider` for the resource key at hand. No provider is registered for `products`, so any workflow transition on a product now fails:

```
No security context provider is registered for resource key "products", so a workflow transition
on it cannot be authorized. Tag a service with
"sulu_content.workflow_transition_request_security_context_provider" and resource-key "products".
Known keys: pages, articles, snippets.
```

### How I ran into it

The failure shows up in SuluMcpBundle's "product bundle" CI job, where publishing a product variant goes through that path: https://github.com/sulu/SuluMcpBundle/pull/49. Five `ProductVariantLifecycleTest` cases fail on a commit that has nothing to do with products, and the same five fail on the untouched base commit once `sulu/sulu` 3.1 is bumped past 09-17, so it is not specific to that PR.

The last green run on this repo's `3.0` is from 09-16, one day before #9100, so the branch simply has not met the new requirement yet.

### Suggested fix

Products carry a fixed context (`ProductAdmin::SECURITY_CONTEXT`), so the ready made `ResourceSecurityContextProvider` covers it. The snippet registration in `SuluSnippetBundle` is the template:

```php
$services->set('sulu_product.workflow_transition_request_security_context_provider')
    ->class(ResourceSecurityContextProvider::class)
    ->args([
        new Reference('doctrine.orm.entity_manager'),
        '%sulu.model.product.class%',
        ProductAdmin::SECURITY_CONTEXT,
    ])
    ->tag('sulu_content.workflow_transition_request_security_context_provider', ['resource-key' => ProductInterface::RESOURCE_KEY])
    ->tag('sulu.context', ['context' => 'admin']);
```

Whether the other securable resources of this bundle (attributes, attribute groups, product families) need one too depends on whether they ever take part in a workflow transition, which you know better than I do.

Happy to open the PR if you want it, and equally happy to leave it to you if you already have this queued. Thanks for the bundle. :slightly_smiling_face:

Amoifr:
Fixes #432. Brings the suite back to green against today's `sulu/sulu` 3.1 dev, in two independent commits.

#### 1. A workflow transition security context provider for products

Since sulu/sulu#9100, every workflow transition is authorized through a provider looked up by resource key. Products have none, so publishing a product, or anything that cascades into a transition such as creating a variant, ends in `UnresolvableSecurityContextException`:

```
No security context provider is registered for resource key "products" [...]
Known keys: pages, examples, snippets.
```

One service fixes it. Products carry a fixed security context (`ProductAdmin::SECURITY_CONTEXT`), so the ready made `ResourceSecurityContextProvider` from the content package covers the case without loading the entity, and the registration mirrors the one `SuluSnippetBundle` does for snippets.

I left the other securable resources of the bundle (attributes, attribute groups, families) alone, since none of them goes through a workflow transition today. Say the word if you would rather have them registered now.

#### 2. Association tests that follow the new resolvable handling

Two more 3.1 changes moved the ground under `ProductResolverAssociationsTest`:

* sulu/sulu#9105 drops resolvables that could not be resolved, so the test that deliberately left its target without published content no longer sees a raw `ResolvableResource`, it sees an empty list;
* sulu/sulu#9102 flattens a resolved entity under a flat output path such as `[product]`, so the values moved from `item['content']['title']` to `item['title']`. That one was motivated by this bundle, #423 built the variant list by hand to work around the old shape.

The targets are published now and the assertions read the flattened item. `testDeclaredFieldParamsAppearInResolvableMetadata` asserted metadata that is no longer observable end to end and is already covered in `ProductSelectionPropertyResolverTest`, so it gives way to a test that pins the new behaviour: an unresolvable target is dropped while its resolvable neighbour in the same product stays, which also keeps the case from passing on an empty list.

#### Tests

No new production behaviour to cover, the existing functional suite could just not pass. On `3.0` with today's 3.1 dev, `Functional` reports **37 failures**, all of them the missing provider, across `ProductControllerTest`, `ProductVariantControllerTest` and `ProductAssociationLifecycleTest`, plus the 3 association ones. With both commits the full suite is green: **1410 tests, 4616 assertions**.

Checked locally on PHP 8.5: `lint-container` OK, php-cs-fixer clean (0 of 367), PHPStan clean. Reverting the association mapping in `ProductResolver` makes all three rewritten tests fail, so they are not passing by accident.


Prokyonn:
Thanks for tracking this down. sulu/sulu#9121 fixes the failure in core: the publish guard no longer
needs a security context for content that no request workflow covers, so products publish again
without a provider.

Could you drop the provider commit from this PR? We'll add the provider once the product bundle
supports request workflows, together with what else that needs, such as `review` in
`ProductAdmin::getSecurityContexts()`. The association test commit can stay here. The flattening
change in `testPublishedTargetResolvesDeclaredProperties` is already on `3.0` through #430, so the
rebase can drop that part.

Amoifr:
Rebased on `3.0`, with the provider commit dropped. The flattening in `testPublishedTargetResolvesDeclaredProperties` now comes from #430, so only the helper and the two realigned tests remain here.

Thanks for sulu/sulu#9121, that is a better place for the fix than a provider the bundle does not need yet. Ping me when the product bundle takes request workflows on, and I will bring the provider back with `review` in `ProductAdmin::getSecurityContexts()`.

For the record, the suite goes from 36 failures to 34 with this commit: it removes `testDeclaredFieldParamsAppearInResolvableMetadata` and `testProductAssociationsAppearUnderRootProductAssociations`, and leaves the rest untouched.

Prokyonn (approved):

Prokyonn:
@Amoifr Thank you!

